PR #83, opened 2018-11-03
#82 - Adds overlay transparency setting

### LINKED ISSUE #82: Feature Request: Transparent overlay setting
To configure whether the overlay should be transparent when not hovered over. The Magic Arena screen is almost entirely used in gameplay for some purpose or the other so the overlay can be a bit intrusive. With transparent setting it can be a lot less cumbersome to use.

farazs:
Fixes #82 

@Manuel-777


farazs:
Screenshots:
transparent_overlay_hover
transparent_overlay

farazs:
Also fixed `undefined cards left` bug

Manuel-777:
This is awesome! My only issue with this right now (and the reason I did not do it before) is that you can't drag the window to the sides of the screen to make it fit perfectly on the sides when its transparent, Something I do very often :P

Manuel-777:
Also, I was thinking maybe instead of opacity use background-color() with alpha to not affect the children and have all children set opacity based on weter or not a setting is on. This would allow card tiles and other elements to 'float' above the game screen

farazs:
> This is awesome! My only issue with this right now (and the reason I did not do it before) is that you can't drag the window to the sides of the screen to make it fit perfectly on the sides when its transparent, Something I do very often :P

I'm not sure I understand. I am able to drag the window to the side of the screen with transparency on. What happens when you do it?

farazs:
> Also, I was thinking maybe instead of opacity use background-color() with alpha to not affect the children and have all children set opacity based on weter or not a setting is on. This would allow card tiles and other elements to 'float' above the game screen

Ah. That seems like a different type of transparency. Only the window itself becomes transparent not the contents? I like the option of having the whole thing be transparent so that it doesn't block my view of the game unless I want to see it. You could add the other option as well with a checkbox perhaps. This is still useful for people who want one type of transparency though right?

farazs:
> > This is awesome! My only issue with this right now (and the reason I did not do it before) is that you can't drag the window to the sides of the screen to make it fit perfectly on the sides when its transparent, Something I do very often :P
> 
> I'm not sure I understand. I am able to drag the window to the side of the screen with transparency on. What happens when you do it?

Oh do you mean the Windows feature to drag to top-right and it auto positions the window? We could make the `transparent` property on the window change based on the alpha setting that would fix this (people would have to choose which is more important).

Manuel-777:
> > > This is awesome! My only issue with this right now (and the reason I did not do it before) is that you can't drag the window to the sides of the screen to make it fit perfectly on the sides when its transparent, Something I do very often :P
> > 
> > 
> > I'm not sure I understand. I am able to drag the window to the side of the screen with transparency on. What happens when you do it?
> 
> Oh do you mean the Windows feature to drag to top-right and it auto positions the window? We could make the `transparent` property on the window change based on the alpha setting that would fix this (people would have to choose which is more important).

Yeah that's what I meant, making it change based on transparency seems a good solution (position will be saved once you set it for the next time it starts anyway)

Also I have been experimenting with adding click trough using this [library](https://github.com/toonvanvr/electron-transparency-mouse-fix), I'll push later if I get it working.

farazs:
@Manuel-777 Updated so transparent property is set based on whether alpha is 1 or not. This requires destroying and recreating overlay if alpha changes from 1 to <1 or vice versa.

For other enhancements like click-through or option to only make the window itself transparent, we can add separately probably?

Separate note: Looks like you have tabs and spaces used interchangeably throughout the code. It sometimes makes the diff look weird on github if some lines are tabbed and some are spaced. Up to you if you want to switch to just one or the other.

Manuel-777:
Made some changes;
I separated the overlay background from its contents into two different wrappers.
I added a new slider on settings to change the background transparency separatedly from the overlay elements.
Moved the transparency sliders to 'Overlay' within settings.
Made elements opacity go to 1 when hovering the ovelay, not the background.

farazs:
Looks better. I like being able to make background fully transparent.

I wonder if there should be a third slider for `controls transparency` that also gets set to 1 on hover but is independent of element transparency. This could include the close minimize etc. icons at the top, the deck name, card counter, and timer at the bottom. I feel like during a game you might want to see just the cards and if you hover, you can see more details. What do you think?

If not, we can probably keep this as is and merge. It's still pretty useful.

Manuel-777:
@farazs I think maybe controls should be limited to the top bar and mode arrows? The only problem is that elements depend on the parent element's alpha. So for example, if we set the 'elements' opacity to 0.5 we cant change the inner elements opacity, we should create a new group of objects for everything that has a different opacity (like I did with elements and background, but keeping in mind the positions)

farazs:
I don't have further changes for now but your latest commit (click-through stuff) broke all the transparency stuff for me. Does it work okay on your end? Maybe we could revert that, merge this and add it back separately after testing.

Manuel-777:
What did it break exactly? It seems to work on my end

farazs:
The overlay is completely opaque (not at all transparent) regardless of setting and cards don't even show up in it anymore. Did you miss checking in changes to package.json? I see error in overlay browser window about `electron-transparency-mouse-fix` being missing even after running npm install.

Manuel-777:
> The overlay is completely opaque (not at all transparent) regardless of setting and cards don't even show up in it anymore. Did you miss checking in changes to package.json? I see error in overlay browser window about `electron-transparency-mouse-fix` being missing even after running npm install.

That was it, also updated other dependencies. Let me know how it works!

farazs:
Works! Okay to merge

farazs:
@Manuel-777 btw you might want to post about this app on https://www.reddit.com/r/MagicArena/. That's where I discovered the other Magic Arena tools and I think a lot of people discover them that way and then tell other people about it.

This one is far better than the others for the in-game overlay and especially draft helper (none of the others have a draft helper).

Unless you don't want too many people to know about it which is totally your call. It may put load on your servers so if you aren't ready for that yet then might be better to wait.

Manuel-777:
I will on the next release, I am not really interested in having tons of users yet, but you are not the first one to point it out! Thanks a bunch for contributing!

PR #104, opened 2018-11-20
Fallback to look for output_log.txt in current directory on non-win32

KerfuffleV2:
This will help non-Windows users since not many people will be running their Wine prefix in the default location. With this change, it's possible to create a link to the `output_log.txt` in the current directory and if the log file does not exist in the default location but does in the CWD then that will be used.

### DIFF
--- a/window_background/background.js
+++ b/window_background/background.js
@@ -545,6 +545,10 @@ if (process.platform === 'win32') {
 else {
     // Path for Wine, could change depending on installation method
     var logUri = process.env.HOME+'/.wine/drive_c/user/'+process.env.USER+'/AppData/LocalLow/Wizards of the Coast/MTGA/output_log.txt';
+    // Fall back to using log file from the current directory if default logUri does not exist but one from CWD does.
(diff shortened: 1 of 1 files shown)


Manuel-777:
I'm thinking If a pop-up indicating the user to enter the URI would be good for this purpose? Then it would be stored on a local file for the next boot of the app.

KerfuffleV2:
> I'm thinking If a pop-up indicating the user to enter the URI would be good for this purpose? Then it would be stored on a local file for the next boot of the app.

I absolutely agree that making it configurable with a file dialog or something would be a lot better than my approach. Unfortunately, that's past my ability to contribute though.

Manuel-777:
Fixed this with PR #107 , thus closing this one.
Check master now, as an input dialog was introduced when log is not in default directory, linux/mac users will have default wine dir checked and if it fails the new dialog will be prompted. Once changed the new direcotry will be saved as default for all users in the same installation.
Cheers!

PR #114, opened 2018-11-26
correct 'wilcard' to 'wildcard'

jmhodges:
Found a few typos in the UI.

There's also a variable name that was typoed, but I can't seem to find it referenced elsewhere. Should I remove it?


jmhodges:
Rebased! If you could take a look, that'd be great!

PR #202, opened 2019-03-01
Added missing fields in package.json.

AnnanFay:
I also changed the author field to match the commit name, but of course change that back if you like the previous value. The contributors are just a list taken from running:

    git shortlog -s -n -e

And removing the obvious duplicates.


Manuel-777:
This is great, thank you man!

PR #205, opened 2019-03-02
Added some eslint warnings and refactored economy.js

AnnanFay:
Added warning to eslint for long functions and high complexity. It's a warning because it's a  goal to aim for.
Refactored economy.js to reduce complexity and split the main function into smaller ones.
Added filter to only show daily summaries in the economy tab.

I've probably broken something somewhere, but it's an improvement over the old code. Let me know if there are any bugs and I will fix them.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -13,6 +13,8 @@
   "rules": {
     "allowEmptyCatch": true,
     "no-cond-assign": 0,
-    "no-console": "off"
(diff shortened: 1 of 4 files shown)


tildar:
Can we push an `eslint ./. --fix` for these changes?

Manuel-777:
It would actually be trough prettier, ill check if it has a command like that

PR #210, opened 2019-03-05
Refactor data

Manuel-777:
Rework of the way we deal with data in the background process.

Basically; create a new object for each match start and store all of that match's relevant data there. Do the same for player data.
This will help in the future to do many things;
- If we want to implement something like React/Angular in the renderer process.
- Better handling of account changes, on a change we erase `playerData` and start over clean We can even keep the old structure for a cleaner swap.
- A centralized point of access for all data is always nice.

Something left to do is to better organize how the historical data is set (economy, history, etc), but as long as they are kept organized its okay. Ideally they should be inside the `playerData` object as well so we can clear those as well on account changes. (which we have to do separately now)


Manuel-777:
Incidentally this brought up some issues with the overlay and the background that were fixed:
- Overlay drafts now work properly, nad we are now able to scroll trough past picks.
- Overlay now differentiates properly between a match and a draft.
- Match data is now centralize in background, so it should be much more consistent and easy to debug.

window_background/labels.js
```diff
@@ -1,1016 +1,1174 @@
...
+              // match end
+              duringMatch = false;
+
+              ipc_send("save_overlay_pos", 1);
+              clear_deck();
+              if (!store.get("settings.show_overlay_always")) {
+                ipc_send("overlay_close", 1);
+              }
+
+              currentMatch.matchId = gameInfo.matchID;
```
dougmill: This specifically subverts a sanity check in `saveMatch`, one that I think I added at some point. `currentMatch.matchId` is expected to be set when the match is created, and if we get a match completed with an id that isn't the same then we've clearly missed something and the data for the match is likely to be incomplete or inaccurate. In such a situation, I think the match should not be saved.

window_background/labels.js
```diff
@@ -1,1014 +1,1180 @@
...
+              currentMatch.game = gameInfo.gameNumber; // probably wrong
```
dougmill: This does not appear to be used for anything. I can't comment on whether it's correct without knowing the intended use.

Manuel-777: I just think its an important piece of data about the game state (just like `currentMatch.bestOf`), might use it in the tourneys api only.

PR #218, opened 2019-03-09
add-electron-unhandled

tildar:
This might be usefull.

https://github.com/sindresorhus/electron-unhandled


Manuel-777:
I am a little worried this might pop tons of errors for users if, say, there is a little exception happening constantly and it goes on a release. I will test for either a way of disabling remotely (via metadata) or add an option to enable/disable these.

tildar:
Agreed... If we can find a way to make it only pop for those of us that are doing development it would be better. I'll look as well.

PR #238, opened 2019-03-28
 Don't throw if an event's match is not found

dougmill:
This was preventing me from opening the events tab, just because one match of a recent event happened to have been played without Tool recording it.

Also, added a flag in match record for running from source.

### DIFF
--- a/window_background/background.js
+++ b/window_background/background.js
@@ -2063,6 +2063,7 @@ function saveMatch(matchId) {
     .getVersion()
     .split(".")
     .reduce((acc, cur) => +acc * 256 + +cur);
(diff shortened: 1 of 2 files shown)


AnnanFay:
This should return `undefined` if there is no found ID. This also fixes all the related code while being consistent.

Returning undefined is useful because you can also use the function to test if a match exists locally - which was the original intention behind having three return cases. I just didn't realise other code relied on missing codes not failing.

```
if(!getMatchesHistroyIndex(key)) {
  // there's no match index so it doesn't exist
}
```

Eh, I've fixed it locally but didn't get round to opening a PR.

PR #248, opened 2019-04-07
Draft replays filter

lusbenjamin:
## New "Draft Replays" Filter on History Tab
draft-replays-filter

originally proposed in discussion with Manwe and Elk
image

### DIFF
--- a/window_main/history.js
+++ b/window_main/history.js
@@ -144,7 +144,7 @@ function open_history_tab(loadMore) {
       select.append('<option value="All">All</option>');
     }
 
-    if (filterEvent != "All Draft") {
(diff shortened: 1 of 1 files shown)


lusbenjamin:
This approach is simple and unlikely to break anything else, but I did wonder if we should eventually try to be "fancier" about always showing the draft replays next to the actual draft matches in some way. This would also make the dynamic, event-specific filters more powerful, for example filtering to "Ranked Draft RNA" would show you just your RNA draft replays alongside your RNA games.

Manuel-777:
Closing , #250 added this already :)

Issue #198, opened 2019-02-26
Missing wildcards count for sideboards


Manuel-777:
Currently cards in the sideboard are counted like an independent object to cards in the mainboard with the same name.

![image](https://user-images.githubusercontent.com/4711391/53426101-1d094c80-39c5-11e9-8984-17bb89c02b62.png)

In the image I own 2 of each glacial fortress and tyrants.

lusbenjamin:
I have a fix for this (mostly) working locally, but it depends heavily on the work in https://github.com/Manuel-777/MTG-Arena-Tool/pull/253

Issue #190, opened 2019-02-19
Incorrect wild cards needed


combataran:
[Take a look](https://i.imgur.com/Zhxp7Pd.jpg) at the sideboard.

I need:
3 duresses
1 negate
1 cast down 
1 kraul harpooner
1 vraska's contempt and
1 hostage taker

The tool says I only need:
1 duress
1 negate and
1 kraul harpooner 

and estimating 3 approximate booster packs to complete the deck. 
My wildcard wheel says I need 7 packs for 2 rare wildcards.

Tool version: latest

Manuel-777:
Make sure that tool is actually reading the log? close and open again. It is possible some cards (like Duress) , that have different versions, get different 'needed' ammounts since Arena does not take that into account while Tool does. But cards like Vraska's Contempt should be accounted perfectly.

combataran:
I renamed the old log, restarted Arena then the tool, so it should be reading fine right?

combataran:
![image](https://user-images.githubusercontent.com/2815112/53059670-a6ff7400-34f2-11e9-88eb-b483bb5f9086.png)
I loaded a new log file and it's now showing me something different.

PR #253, opened 2019-04-08
Wildcards and Estimated Booster Stats by Set

lusbenjamin:
## Motivation
- Given the partially complete decks I have on Arena, which booster pack should I buy next?
  - How many cards do I want in each set? How many do I want by rarity?
  - How many booster packs in each set would I need to complete all my decks?
- ~~How many booster packs in each set would I need to complete the entire set?~~
  - This math is much harder than estimations based on wildcards or duplicate protection only. Also, this probably isn't useful.
- ~~How many booster packs in each set would I need to complete the entire set of rares/mythics?~~
  - ~~This math is simple because of duplicate protection and more useful for deck building or knowing when you will only get gems from packs.~~
  - This math is also hard. 😖 
- For the next booster pack I buy in each set, what is the chance that the rare or mythic would help me complete my decks? 💯 ✅ 2️⃣ ➕ 2️⃣ ✅  

## Demo
image

## Other Thoughts
- This is probably closely related to https://github.com/Manuel-777/MTG-Arena-Tool/issues/198 and https://github.com/Manuel-777/MTG-Arena-Tool/issues/190 but hopefully only tangential. Those bugs probably relate to `util.get_deck_missing`, which this PR avoids.
- I have yet to set up any kind of performance-related profiling. Empirical tests on my machine seem to show that this doesn't add much additional load time to the stats panel, but an increase in the number of decks might require us to push this into background computation.

### DIFF
--- a/window_main/collection.js
+++ b/window_main/collection.js
@@ -38,10 +38,10 @@ function openCollectionTab() {
 
   orderedSets.sort((a, b) => {
     if (setsList[a].release < setsList[b].release) {
-      return -1;
+      return 1;
     }
     if (setsList[a].release > setsList[b].release) {
-      return 1;
+      return -1;
     }
     return 0;
   });
--- a/window_main/collection.js
+++ b/window_main/collection.js
@@ -498,27 +498,43 @@ function renderSetStats(setStats, setIconCode, setName) {
     });
 
     // If the set has a collationId, it means boosters for it exists
-    if (setsList[setName].collation) {
-      let wantedDiv = createDivision(["stats_set_completion"]);
-      let wantedCost = getBoosterCountEstimate(wanted);
-      let wantedIcon = createDivision(["stats_set_icon", "bo_explore_cost"]);
-      wantedIcon.style.height = "30px";
-      let wantedSpan = document.createElement("span");
-      wantedSpan.innerHTML = `<i>~${wantedCost} boosters to craft wanted cards using wilds.</i>`;
-      wantedIcon.appendChild(wantedSpan);
-      wantedDiv.appendChild(wantedIcon);
-      substats.appendChild(wantedDiv);
+    if (setsList[setName] && setsList[setName].collation) {
+      let chanceBoosterHasMythic = 0.125; // assume 1/8 of packs have a mythic
+      let chanceBoosterHasRare = 1 - chanceBoosterHasMythic;
+      let wantedText =
+        "<abbr title='missing copy of a card in a current deck'>wanted</abbr>";
 
-      // This is exact because we can depend on duplicate protection
-      let rareMythicCompletion = missing["rare"] + missing["mythic"];
-      let completionDiv = createDivision(["stats_set_completion"]);
-      let completionIcon = createDivision(["stats_set_icon", "bo_explore_cost"]);
-      completionIcon.style.height = "30px";
-      let completionSpan = document.createElement("span");
-      completionSpan.innerHTML = `<i>${rareMythicCompletion} boosters to open all rares and mythics.</i>`;
-      completionIcon.appendChild(completionSpan);
-      completionDiv.appendChild(completionIcon);
-      substats.appendChild(completionDiv);
+      // chance that the next booster opened contains a rare missing from one of our decks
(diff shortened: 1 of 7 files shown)


Manuel-777:
Some sets need to be filtered from these calculations, since boosters for those do not exists, like Mythic Edition and M19 Gift Pack.
I could add a flag in the metadata setsList for sets that are 'acquirable' or you can buy boosters from.

lusbenjamin:
Now that I understand that our booster estimation math was purely based on producing wildcards, I think we should take out the "~X estimated boosters for all cards" line. No one would actually complete a collection using purely wildcards, and if they buy a bunch of boosters in that set the math is different.

window_main/collection.js
```diff
@@ -490,7 +493,41 @@ function renderSetStats(setStats, setIconCode, setName) {
...
+    let wantedIcon = createDivision(["stats_set_icon", "bo_explore_cost"]);
+    wantedIcon.style.height = "30px";
+    let wantedSpan = document.createElement("span");
+    wantedSpan.innerHTML = `<i>~${wantedCost} estimated boosters for wanted cards.</i>`;
+    wantedIcon.appendChild(wantedSpan);
+    wantedDiv.appendChild(wantedIcon);
+    substats.appendChild(wantedDiv);
+
+    // This is very vague, assuming the distribution of mythic cards is 7 rares out of 8 packs.
+    let estimatedRareCompletion = Math.round(missing["rare"] / 0.875);
```
lusbenjamin: Given that I am still missing X rares, I think this answers the question "On average, how many boosters do I need to open to get X rares?" but does not answer the question "On average, how many boosters do I need to open to get the specific rares that I still need?"

Manuel-777: Yes, it was not meant to answer that question specifically, for that the chance is 1/X where X is the number of missing rares, thus a hypergeometric distribution would suit for the calculation.

lusbenjamin: I was originally hoping to use our `util.hypergeometricSignificance` code to answer the second question, but I think that may be incorrect. Although I am fuzzy in stats, I believe that we need a binomial distribution instead of a hypergeometric distribution because the act of opening a *digital* booster does not "remove any cards from the population", in other words, your chances of getting a Krasis are the same regardless of whether or not the last pack you opened contained a Krasis.

Manuel-777: It does remove a chance, since there is duplicate protection now you cant open more than 4x of every rare and mythic

lusbenjamin: (I am also totally okay on punting this feature if it is too hard, since I'm not sure that it will be valuable to many users)

lusbenjamin: Oh man, I totally forgot about duplicate protection. Doesn't that make the math even harder?

lusbenjamin: Halp @AnnanFay, too many maths! 😖

lusbenjamin: ...wait. Duplicate protection does make the math super simple and exact. I'm being dumb.

Manuel-777: It's easier actually because we just count as a normal hypergeometric, with pop size, number of occurrences and sample size afaik

lusbenjamin: Since any duplicates are guaranteed to be cards that you are missing, the answers to the two questions actually are exactly the same!

lusbenjamin: (aka, your math as written is correct and I'm finally understanding it.)

Manuel-777: @dougmill is more experienced in stats trough, he will know better :)

lusbenjamin:
@Manuel-777 g2g!

lusbenjamin:
Cancel that g2g. 😕 

Going to try again to clean this up in a bit:
image

lusbenjamin:
image
- realigns numbers to be pseudo-tabular for easy scanning
- adds hover text to give some explanation of 'wanted cards'
- removes wc estimation metric
- removes former completion estimation metrics
- adds 2 new metrics based on AnnanFay maths

lusbenjamin:
@AnnanFay, will you please double check that I accurately implemented the math we discussed? @Manuel-777 and @AnnanFay any other feedback welcome.

PR #258, opened 2019-04-13
Auto login and Launch to tray

lusbenjamin:
## Motivation
image

## Approach
- Refactor signals and handlers related to settings and config
  - `background` handles config persistence and resolution
    - new `rstore.settings` holds app config, new IPC signals and handlers for saving
    - all handlers that modify stores call `loadSettings` afterwards to propagate changes
      - eliminated `relay` concept entirely
    - `loadSettings` reads defaults, then user config, then app config
      - passes blended config on to IPC
  - `main`, `renderer`, and other windows can continue to use a unified `settings` downstream of IPC
- Add `auto_login` and `launch_to_tray` to app config
- Add checkboxes and new `setAppSettings` handler to Settings page
  - Launch to tray is only possible if Auto login is enabled
  - [ ] TODO make Auto login checkbox appear disabled when necessary
- Refactor startup bootstrapping logic
  - Display loading screen until _after_ `set_db` signal
  - Send new `auto_login` signal if enabled
  - Hide window if `launch_to_tray` enabled
    - Display window if any issue occurs in auth
  - Otherwise activate login form and show as before

## Other Refactors
- extract magic pw string "********" to `util.HIDDEN_PW`
- move "Remember Me" state into `rstore.settings`
- future config settings will auto-populate defaults (see comment below)

## Bonus Bugfixes
- Gracefully handle opening the History page with no data
- [x] Correctly move to offline mode when "Privacy > Online Sharing" is disabled

### Auto Login Online
auto-login-online

### Launch to Tray
launch-to-tray

### Auto Login Offline
auto-login-offline

### DIFF
--- a/main.js
+++ b/main.js
@@ -176,7 +176,7 @@ function startApp() {
 
       case "set_settings":
         //console.log("set settings: ", arg);
-        saveSettings(arg);
+        setSettings(arg);
         mainWindow.webContents.send("set_settings", arg);
         overlay.webContents.send("set_settings", arg);
         break;
@@ -201,12 +201,6 @@ function startApp() {
       case "set_cards":
         mainWindow.webContents.send("set_cards", arg.cards, arg.new);
         overlay.webContents.send("set_cards", arg.cards);
-        break;
-
-      case "save_settings":
-        saveSettings(arg);
-        background.webContents.send("save_settings", arg);
-        overlay.webContents.send("set_settings", arg);
         break;
 
       case "renderer_update_install":
@@ -382,7 +376,7 @@ function startApp() {
   });
 }
 
-function saveSettings(settings) {
+function setSettings(settings) {
   app.setLoginItemSettings({
     openAtLogin: settings.startup
   });
--- a/main.js
+++ b/main.js
@@ -36,6 +36,7 @@ const ipc = electron.ipcMain;
 
 var mainLoaded = false;
 var backLoaded = false;
+var firstSettingsRead = true;
 
 //commandLine, workingDirectory
 const singleLock = app.requestSingleInstanceLock();
@@ -132,7 +133,6 @@ function startApp() {
     mainLoaded = true;
     if (backLoaded == true) {
       background.webContents.send("set_renderer_state", 1);
-      showWindow();
     }
   });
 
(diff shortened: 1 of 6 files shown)


window_background/background.js
```diff
@@ -858,39 +888,31 @@ function syncUserData(data) {
...
-  //console.log(_settings);
-  //const exeName = path.basename(process.execPath);
+// Loads and combines settings variables, sends result to display
+function loadSettings() {
+  // Blends together default, user, and app config
+  // Since settings have migrated between areas, collisions happen
+  // Order of precedence is: app > user > defaults
+  const settings = store.get("settings");
+  const rSettings = rstore.get("settings");
+  const _settings = { ...defaultCfg.settings, ...settings, ...rSettings };
```
lusbenjamin: This cool snippet ensures that new config settings will always take their defaults from `defaultCfg.settings` without needing us to remember custom logic to add defaults here.

main.js
```diff
@@ -203,12 +211,6 @@ function startApp() {
         overlay.webContents.send("set_cards", arg.cards);
         break;
 
-      case "save_settings":
-        saveSettings(arg);
-        background.webContents.send("save_settings", arg);
-        overlay.webContents.send("set_settings", arg);
-        break;
```
lusbenjamin: Too confusing to have partially updated (dirty) settings state passing in signals in both directions. I eliminated this one to force all changes to go through the logic in `background.loadSettings`.

main.js
```diff
@@ -382,11 +384,13 @@ function startApp() {
   });
 }
 
-function saveSettings(settings) {
+function setSettings(settings) {
```
lusbenjamin: a minor rename for clarity. I tried to use "save" only when persisting changes to stores and "set" when updating display state.

window_background/background.js
```diff
@@ -1256,7 +1278,6 @@ function processLogUser(rawString) {
     if (value.indexOf(strCheck) > -1) {
       playerData.name = dataChop(value, strCheck, '"');
       ipc_send("set_player_data", playerData);
-      ipc_send("init_login", true);
```
lusbenjamin: This was displaying the login form before it was possible to use (before `"set_db"`).

PR #259, opened 2019-04-15
clean up history filters

lusbenjamin:
- hide practice matches from format filter ("AIBotMatch")
- show deleted deck names in deck filter
- move deleted decks to end of deck filter

## Motivation
image

## Demo
filterpatch

### DIFF
--- a/shared/util.js
+++ b/shared/util.js
@@ -1147,37 +1147,6 @@ function getReadableEvent(arg) {
 }
 
 //
-function getReadableDeckName(deck_id) {
-  if (typeof decks === 'undefined') {
-    return deck_id;
-  }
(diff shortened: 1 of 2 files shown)


lusbenjamin:
When I first built this, I was a little concerned about a vanilla dropdown potentially having to handle hundreds of decks. I dismissed the fear at the time because I mistakenly thought we were protected by the 60 deck limit in Arena. I forget to take into account draft decks and deleted decks! Not sure this will break immediately, but this will definitely eventually break as users build up history over years.

Manuel-777:
I think we should really keep showin these, but instead of hiding fix the name/cost. Maybe create a temporary list of ID to name/cost on history load?

PR #264, opened 2019-04-17
Add batchfile

JamesYouL2:
This is not amazing coding, but I think it's probably useful. It's a batch file that allows you to start MTGA Tool by double clicking instead of running from bash shell. I think it's good because you can put it as a shortcut on your desktop or whatever.

### DIFF
--- a/npmstart.bat
+++ b/npmstart.bat
@@ -1,0 +1,2 @@
+echo before
+npm start


Manuel-777:
I added a build entry to sublime text, so I press Ctrl+B and it runs using sublime's console.

What are you using to code? I think VS has that too!

JamesYouL2:
VS Code. Not really used to it yet, but yeah, I got it figured out. Thanks.

PR #274, opened 2019-04-22
collection filter rarity UI change

ptasznikarium:
before:
image

after:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ main.sublime-project
 *.sublime-settings
 .vscode
 .DS_Store
+/.vs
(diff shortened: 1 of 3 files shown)


Manuel-777:
Is there a good reasoning for this change? I remember it being this way before but a collab- UI Designer recommended our current styles

ptasznikarium:
Goal was to add option for showing only incomplete set cards.
During that I thought it would be nicer to have rarity filter that way and it would free up some space for extra option
I was not aware of it being that way before and got changed
feel free to reject it, no emotional attachment

PR #279, opened 2019-04-23
add 'Incomplete only' filter in Collection

ptasznikarium:
New advanced filter in Collection to show only cards without complete playset (less than 4 copies owned)

collection-filter-playset-2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ main.sublime-project
 *.sublime-settings
 .vscode
 .DS_Store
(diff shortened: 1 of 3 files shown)


ptasznikarium:
some shenanigans with icon, but did not wanted to create any image in case you want different icon

Manuel-777:
Icon is good! I could make an image just like it but its the same.
if you have any better icon ideas I can draw it tho

PR #280, opened 2019-04-23
remove obsolete flag in Collection filters

ptasznikarium:
Micro optimization of code by removing obsolete flag and replacing it iteration skips


ptasznikarium:
JS forEach is so evil when you want to use _break_, _continue_ or _return_ ....

ptasznikarium:
disregard this, will do it tomorrow in cleaner way

Issue #283, opened 2019-04-25
Error Attempting to Open Partial Draft Replay


lusbenjamin:
Related to https://github.com/Manuel-777/MTG-Arena-Tool/issues/282, if a player stops a draft in the middle of pack-pick selection, the Tool may lose some of the draft history necessary to display the replay on the History page. Attempting to click on one of these partially corrupt replays will silently throw a JS error.

PR #286, opened 2019-04-26
Backports and Bugfixes

lusbenjamin:
## Random Backports and Bugfixes
This is a handful of backports from the current build in https://github.com/lusbenjamin/MTG-Arena-Tool/tree/create-react-app

### Overlay
  - Added a "Current" to the title for the current pack/pick in draft mode.
  - I believe the default for maximum pack size should be `15 - 1 = 14`, not sure why it was `13` before
  - In some odd edge cases, the draft mode next pick toggle can go beyond the 3rd pack. I fixed the boundary logic to catch those cases.
  - When the main content requires scroll, the final bit can be blocked by the floating clock. I added a spacing `div` before the clock to fix this.
  - Implements sorting and deduplication using the new object utilities. ([we had a feature regression here](https://github.com/lusbenjamin/MTG-Arena-Tool/commit/89a29572a81e73152b8666712fe0b0fa2c0b4f26#diff-1435c8c1268f77062c3d0894c22de331L398))
image

### Action Log
  - ~~Added a stabilizing `defaultIndex` to keep log entries with duplicate times from jumping around~~
  - ~~Reversed log order so recent entries always appear at the top~~

### Economy Page
  - Opening your vault will now only display your positive Vault progress in the daily summary bar
image

### History Page
  - Bugfix for https://github.com/Manuel-777/MTG-Arena-Tool/issues/283 - Opening a draft replay with partially complete pack/pick data will now fail gracefully

### DIFF
--- a/window_main/economy.js
+++ b/window_main/economy.js
@@ -232,7 +232,11 @@ function createDayHeader(change) {
   const icva = tx.cloneNode(true);
   icva.innerHTML = "Vault:";
   const vatx = tx.cloneNode(true);
-  const deltaPercent = dayList[daysago].vaultProgress / 100.0;
+  const rawDelta = dayList[daysago].vaultProgress;
+  // Assume vault can only be redeemed once per day
+  // Rely on modulo arithmetic to derive pure vault gain
+  const delta = rawDelta < 0 ? rawDelta + 100 : rawDelta;
+  const deltaPercent = delta / 100.0;
   vatx.innerHTML = deltaPercent.toLocaleString([], {
     style: "percent",
     maximumSignificantDigits: 2
--- a/shared/cards-list.js
+++ b/shared/cards-list.js
@@ -15,6 +15,7 @@ class CardsList {
     if (!list) list = [];
     this._list = list.map(obj => {
       return {
+        ...obj,
         quantity: obj.quantity || 1,
         id: obj.id || obj,
         measurable: true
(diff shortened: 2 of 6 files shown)


window_main/economy.js
```diff
@@ -232,7 +232,11 @@ function createDayHeader(change) {
   const icva = tx.cloneNode(true);
   icva.innerHTML = "Vault:";
   const vatx = tx.cloneNode(true);
-  const deltaPercent = dayList[daysago].vaultProgress / 100.0;
+  const rawDelta = dayList[daysago].vaultProgress;
+  // Assume vault can only be redeemed once per day
+  // Rely on modulo arithmetic to derive pure vault gain
+  const delta = rawDelta < 0 ? rawDelta + 100 : rawDelta;
```
lusbenjamin: It may be a better idea to have the back end send separate vault gain and vault loss, but this hack is simple and definitely good enough for 99.9% of non-whale players.

lusbenjamin:
Hmmm. I didn't notice https://github.com/Manuel-777/MTG-Arena-Tool/commit/7f996edff762d5c97bde750d09b7bf657fb5fb9d happened on `master` after `create-react-app` forked, so I never made any attempt to build "auto-scroll" into the React version. Reverse order seems like a better user experience to me, but I am open to other thoughts.

Manuel-777:
> Hmmm. I didn't notice [7f996ed](https://github.com/Manuel-777/MTG-Arena-Tool/commit/7f996edff762d5c97bde750d09b7bf657fb5fb9d) happened on `master` after `create-react-app` forked, so I never made any attempt to build "auto-scroll" into the React version. Reverse order seems like a better user experience to me, but I am open to other thoughts.

Yep, I would revert that, I think it makes more sense for the action log to grow underneath like a chat, instead of going up.

lusbenjamin:
Added overlay sorting and deduplication. I also backported my solution for copying the chance fields to avoid copying over the `_list` field manually.

PR #292, opened 2019-04-28
crop deck images higher up

lusbenjamin:
kaya-before
kaya-after

Looks better for most images:
image


lusbenjamin:
For @AnnanFay ;)
image

Manuel-777:
EEEHH, I reckon it looks mostly good, but it depends on the art. WIll merge but I will keep an eye open!

Issue #296, opened 2019-04-28
Cards left in deck overlay no longer shows correct information with newest update


ianelletson:
![image](https://user-images.githubusercontent.com/1207114/56867541-f0b18280-69ab-11e9-9a59-13f99c0d434f.png)
Percentages do not add up to 100%

![image](https://user-images.githubusercontent.com/1207114/56867545-fb6c1780-69ab-11e9-8601-8859f90c4fcf.png)
Not ordered by percent remaining descending

lusbenjamin:
I can easily reproduce the issues with the percentages. 🤔 we may have had a feature regression in the Sample Size panel when we [migrated to the new interpreter engine](https://github.com/Manuel-777/MTG-Arena-Tool/commit/89a29572a81e73152b8666712fe0b0fa2c0b4f26#diff-9179d37aa838679c9dd3c54d801755f9).

The second issue with ordering by percent was fixed in https://github.com/Manuel-777/MTG-Arena-Tool/pull/286 and should ship with the next release.

Manuel-777:
I can reproduce too, definitely broke the calculations on last update, will fix asap

ianelletson:
Thank you very much for your hard work.

PR #298, opened 2019-04-29
bugfix sample size math

lusbenjamin:
## Motivation
We had a regression in our Sample Size panel when we [migrated to the new interpreter engine](https://github.com/Manuel-777/MTG-Arena-Tool/commit/89a29572a81e73152b8666712fe0b0fa2c0b4f26#diff-9179d37aa838679c9dd3c54d801755f9). I believe our current math always assumes you have exactly `odds_sample_size` (e.g. 1) creature/land/etc instead of using the count in the deck.

Fixes the other half of https://github.com/Manuel-777/MTG-Arena-Tool/issues/296

## Approach
Revert to old correct math. ~~Refactor to use a single helper method while I'm at it.~~

## Demo
Annotation 2019-04-29 132658

### DIFF
--- a/window_background/background.js
+++ b/window_background/background.js
@@ -1662,40 +1662,32 @@ function forceDeckUpdate(removeUsed = true) {
     }
     let main = currentMatch.playerCardsLeft.mainboard;
     main.addProperty("chance", card =>
-      Math.round(
-        hypergeometricRange(
-          1,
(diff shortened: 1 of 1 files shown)


lusbenjamin:
Oh shoot, didn't see that weird rounding detail. Reverting the refactor part of this fix.

lusbenjamin:
Restored extra significant figure:
image

PR #305, opened 2019-05-02
Potential Bugfix for JS error on Overlay load

lusbenjamin:
I don't have any experience with custom elements, and I'm not sure what the "correct" approach should be. This is just the first working solution I could produce.

image

### DIFF
--- a/window_overlay/overlay.js
+++ b/window_overlay/overlay.js
@@ -916,10 +916,10 @@ class LogCard extends HTMLElement {
 function addLogProtos() {
   const logCardElement = window.customElements.get("log-card");
   if (!logCardElement) {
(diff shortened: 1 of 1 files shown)


Manuel-777:
[1d21f2](https://github.com/Manuel-777/MTG-Arena-Tool/commit/1d21f27574b3083f5819d5998317147a7f80bd01) Explanation;
I think Jquery on append()/html() handles the custom elements already as normal HTML elements, so there was no need to add them manually, I dont know if this is "hacky", but it seems to work and is far less complex.
When we move out of jquery onto react I hope we can keep the custom tags as react components and add them the same way we do now :)

PR #310, opened 2019-05-04
bugfix decks page drilldown when filters active

lusbenjamin:
### Motivation
When a filter is active, clicking on a deck will open up the incorrect deck details page. (You will see the Nth deck as if you clicked in the same location with no filters active.) I believe I introduced this bug back when I put filters on the page. 🤕 

This bugfix makes the page have the correct behavior.

### DIFF
--- a/window_main/decks.js
+++ b/window_main/decks.js
@@ -23,7 +23,6 @@ global
   StatsPanel
 */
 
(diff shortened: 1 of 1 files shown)


Manuel-777:
Saw this one and was going to try to fix it, glad you did it first :P

PR #313, opened 2019-05-04
reverse log order

lusbenjamin:
### Motivation
Our first load of the History page causes an expensive "worst-case scenario" sort of the matches history, flipping its order so that it can be displayed most-recent-first in the page. This optimization changes the final  ordering to closer to "in place" or "best-case scenario" and then tweaks the display code to iterate over it in reverse order instead.

Typical chart from performance recording of initial load of History page from clean start up:
image

### DIFF
--- a/window_main/history.js
+++ b/window_main/history.js
@@ -145,7 +145,8 @@ function open_history_tab(loadMore) {
     actuallyLoaded - begin < filteredSampleSize;
     loadHistory++
   ) {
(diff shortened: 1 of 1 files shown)


lusbenjamin:
This has other consequences to logic downstream, where other things depended on the global `matchesHistory` being sorted in reverse order. I'm closing this for now until I think this through.

PR #318, opened 2019-05-06
Panel resize listeners

lusbenjamin:
### Demo
the flicker of loading bar you see is because it saves it to user settings to persist across sessions
panel-resize-listeners


lusbenjamin:
### Motivation
image

PR #321, opened 2019-05-07
bugfix settings persistence

lusbenjamin:
### Motivation
Had to save @Manuel-777 from the recurring Lavarunner.
image

### Demo of Fix
settings-bugfix


lusbenjamin:
this is not the fix I thought it was. I need to think this through a lot more thoroughly.

PR #325, opened 2019-05-07
Remove allmatches singleton

lusbenjamin:
### Motivation
image

### Approach
This removes the global singleton aggregation of all match data `allMatches` that snuck its way into https://github.com/Manuel-777/MTG-Arena-Tool/pull/320. The History page has reverted back to maintaining its own copy, and now the Decks page maintains one as well (necessary for the event filter dropdown).

### DIFF
--- a/window_main/decks.js
+++ b/window_main/decks.js
@@ -1,7 +1,6 @@
 /*
 global
   Aggregator,
-  allMatches,
 	cardsDb,
   createDivision,
 	decks,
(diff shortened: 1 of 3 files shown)


lusbenjamin:
Not clear that we need or want this yet. [I hardened the `Aggregator.compareDecks` and `Aggregator.compareEvents` functions just in case.](https://github.com/Manuel-777/MTG-Arena-Tool/pull/323/commits/c32bdb4e4ee9384bc590331649cb389148e315b5)

lusbenjamin:
https://discordapp.com/channels/463844727654187020/506793351786266624/575393676625313812

Current best guess is that the original problem is due to the fact that `master` instantiates an `Aggregator` and processes `matchesHistory` before the History page `history.sort_history` has a chance to fill in corrupted matches with fake deck data. https://github.com/lusbenjamin/MTG-Arena-Tool/blob/master/window_main/history.js#L866

lusbenjamin:
closing this and hardening the `Aggregator` in #323 to make it handle `matchesHistory` without relying on `history.sort_history`. We can re-open if it turns out we need to.

PR #330, opened 2019-05-10
Travis Eslint

lusbenjamin:
### Motivation
- The ultimate goal here is help us get more value out of Eslint by making it part of our build process.
- Ideally certain types of linting errors that are high-risk can be used to help us keep bugs from going to production.
- We want to be able to have the local build exactly match the build Travis does to mark a success

### Approach
- Have the build use local project `eslint` and `.eslintrc.json` settings
- Ensure `eslint` is called before building
- Take all current lint errors and make them warnings
- If we decide they are valuable to maintain in the future, we can fix those warnings and convert the `.eslintrc.json` setting back to an "error"


lusbenjamin:
Nice! Failing there is exactly what we want. Please wait to merge until I get this to pass.

Manuel-777:
Good travis!

main.js
```diff
@@ -142,6 +142,12 @@ function startApp() {
     updaterWindow = undefined;
   }
 
+  //
+  let updateState = null;
+  ipc.on("set_update_state", function(event, arg) {
+    updateState = arg;
+  });
```
lusbenjamin: @Manuel-777 please double check this change for me, it is the one I am least confident that I fully understand. How is `master` working right now with this signal? I do not understand the auto updater yet.

Manuel-777: Why was this added? Update state should be sent from main to renderer in any case

lusbenjamin: I thought that `main` and `renderer` were separate processes and different window scope? Further down in `main`, we need `updateState` to make a dispatch of some kind:
```
case "renderer_update_install":
        if (updateState == 3) {
          autoUpdater.quitAndInstall();
        }
        background.webContents.send("update_install", 1);
        break;
```
However `updateState` only lives in the main window process scope?

lusbenjamin: The additional declared variable and handler I added here is an attempt to get the same variable into the `main` process. But I don't really know what's supposed to happen here.

lusbenjamin:
The build is currently broken because it attempts to merge into `master`, and `master` has some broken code in `deck_details.js`:
image

lusbenjamin:
Since waiting for Travis to tell you that things are broken is slow and a little annoying, here are several ways to get earlier feedback before pushing (from slowest to fastest):
1. Run `npm run dist`, since you should always make sure the build works.
2. Run `npm test`. This is now an alias for the full lint run that we do before `npm run dist`.
3. I added a new git pre-push validation script. Optionally run `./git-pre-push.sh` and it will install a pre-push hook in your repository. From then on, git will run that hook and only push commits to the remote if they pass. It will only lint the files that are different since the remotely tracked branch. Attempts to push broken commits will fail, so this is also the most opinionated approach!

PR #333, opened 2019-05-13
Travis jest

lusbenjamin:
### Motivation
- restore our tests in `window_background/arena-log-decoder/__tests__/`
- hook our tests up to our build process
  - tests must all pass as a validation check before `npm run dist`
  - tests must all pass for TravisCI build to pass
  - all test-related code and `output_log.txt` must be excluded from build
### Approach
- restore `jest` as dev dependency
- add current `output_log.txt` and get tests working again
- update `package.json` script endpoints to call `jest` during build

### Demo

```
$ npm run lint

> MTG-Arena-Tool@2.5.2 lint C:\Users\Local Ben\code\MTG-Arena-Tool
> npx eslint . --quiet
```
```
$ npm test

> MTG-Arena-Tool@2.5.2 pretest C:\Users\Local Ben\code\MTG-Arena-Tool
> npm run lint


> MTG-Arena-Tool@2.5.2 lint C:\Users\Local Ben\code\MTG-Arena-Tool
> npx eslint . --quiet


> MTG-Arena-Tool@2.5.2 test C:\Users\Local Ben\code\MTG-Arena-Tool
> npx jest

PASS window_background/arena-log-decoder/__tests__/arena-log-decoder-spec.js
PASS window_background/arena-log-decoder/__tests__/nth-last-index-spec.js
PASS window_background/arena-log-decoder/__tests__/json-text-spec.js

Test Suites: 3 passed, 3 total
Tests:       9 passed, 9 total
Snapshots:   0 total
Time:        1.543s
Ran all test suites.
```
```
$ npm run dist

> MTG-Arena-Tool@2.5.2 predist C:\Users\Local Ben\code\MTG-Arena-Tool
> npm test


> MTG-Arena-Tool@2.5.2 pretest C:\Users\Local Ben\code\MTG-Arena-Tool
> npm run lint


> MTG-Arena-Tool@2.5.2 lint C:\Users\Local Ben\code\MTG-Arena-Tool
> npx eslint . --quiet


> MTG-Arena-Tool@2.5.2 test C:\Users\Local Ben\code\MTG-Arena-Tool
> npx jest

PASS window_background/arena-log-decoder/__tests__/arena-log-decoder-spec.js
PASS window_background/arena-log-decoder/__tests__/json-text-spec.js
PASS window_background/arena-log-decoder/__tests__/nth-last-index-spec.js

Test Suites: 3 passed, 3 total
Tests:       9 passed, 9 total
Snapshots:   0 total
Time:        1.562s
[...]

PR #334, opened 2019-05-13
CI: Validate source formatting with prettier

gavacho:


gavacho:
This will configure travis-ci to test that the source code has been correctly formatted with prettier.

To merge this, it looks like we'd have to:
1) Configure prettier to ignore files we don't author (like jquery) -- easiest way to do this would be to move them into a directory called `vendor` or something.
2) Use prettier to format the unformatted files

lusbenjamin:
I like that standardized formatting makes it faster to understand shared code. I dislike losing time to [bike-shedding](https://en.wikipedia.org/wiki/Law_of_triviality). I lean slightly to some kind of approach that uses prettier `--write` automatically because I've gotten used to relying on transpiling source code anyway.

lusbenjamin:
I like @gavacho's more precise `npm` scripts, but note that this will conflict with #333

gavacho:
I've rebased of master and addressed the conflicts

package.json
```diff
@@ -51,10 +51,11 @@
   "scripts": {
     "start": "electron .",
     "postinstall": "install-app-deps",
-    "lint": "npx eslint . --quiet",
-    "pretest": "npm run lint",
-    "test": "npx jest",
-    "predist": "npm test",
+    "format": "prettier --write \"**/*.js\"",
+    "test": "npm run --silent test:eslint && npm run --silent test:prettier && npm run --siltent test:jest",
```
lusbenjamin: `npm run --siltent test:jest` ==> `npm run --silent test:jest`

gavacho: thanks for catching this!

lusbenjamin:
🤣 can confirm that PR works as intended. build only failed because we merged immediately after first build before 2nd could succeed.

PR #338, opened 2019-05-14
Husky Git Hooks

lusbenjamin:
### Motivation
- Committing to the project should automatically run `prettier`
- Attempting to push to a remote repo should pre-validate with `npm test`

### Approach
This follows the "Option 1" recommendation here: https://prettier.io/docs/en/precommit.html
- Adds dev dependency on https://github.com/typicode/husky to do nice git hooks
- Adds dev dependency on https://github.com/okonet/lint-staged to auto-format pre-commit
- Remove old pre-push hook

### Note
This is a rather opinionated PR, so we should get consensus from contributors before merging.


lusbenjamin:
@gavacho mentioned https://github.com/ghooks-org/ghooks as an alternative approach. I did not evaluate it against the `husky + lint-staged` approach

lusbenjamin:
The auto-formatting is pretty sweet, but it takes just long enough to make GitHub Desktop feel unresponsive at times. No problems using it while in terminals because of the obvious output.
I've learned you can opt out of them entirely by simply having your own `pre-push` and `pre-commit` scripts already installed. Husky will not overwrite them, even if they do nothing and effectively bypass the project recommendation
and then on the command line there is always `git push --no-verify`
(which still doesn't count as a `--force` :wink:)

lusbenjamin:
closing this in favor of opt-in git hooks https://github.com/Manuel-777/MTG-Arena-Tool/pull/364

PR #339, opened 2019-05-14
Archive Decks and Events

lusbenjamin:
### Motivation
Archive filter had a cousin with amnesia on the Decks page? There was a parallel universe where archive filter was also on the Events page? Moar sequels in the expanding Filters Multiverse.

### Demo
events-archive
decks-archive

### DIFF
--- a/shared/Aggregator.js
+++ b/shared/Aggregator.js
@@ -158,8 +158,11 @@ class Aggregator {
     if (!passesDeckFilter) return false;
 
     const currentDeck = getDeck(deck.id);
-    if (!showArchived && currentDeck && currentDeck.archived) return false;
-    
+    const passesArchiveFilter =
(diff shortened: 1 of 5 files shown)


Manuel-777:
I heard this guy was rather evil, and instead of archiving stuff he deletes them! :O

lusbenjamin:
These last few commits were thrashing as I tested that archived state and logic worked correctly between the decks page and the events page. Everything behaves as I currently expect:
- archiving a deck affects the summary stats
- archiving a deck does not hide matches from history page
- BUT archiving a match on the history page affects the stats on the deck page
- enabling the archive filter always includes all data
- **Edge case** expanding an event on the Events page always shows all matches, regardless of archive state

Manuel-777:
It says aggregator has conflicts to merge but I cant see them?

PR #343, opened 2019-05-14
Filter Consistency Bugfixes

lusbenjamin:
- Makes the archive checkbox independent on the Decks and History pages to match the Events and Economy pages
- Updates the filter logic on the Events page so that individual match.archived state is reflected in event summary stats
- Updates the filter logic on the Events page so that opening an event respects individual match.archived state
- Showing archived always shows all matches


lusbenjamin:
economy-archive-filter

PR #350, opened 2019-05-16
Ranked Stats

lusbenjamin:
### Motivation
https://discordapp.com/channels/463844727654187020/467737642306371584/578131652136534027
image

Refactors the stats panel:
- removes redundant stats
- groups winrates together

### Before / After
image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,7 +20,7 @@
     "no-undef": "error",
     "no-global-assign": "warn",
     "no-empty": "warn",
-    "complexity": ["warn", 20],
+    "complexity": ["warn", 40],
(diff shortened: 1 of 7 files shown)


lusbenjamin:
Limited season now uses limited rank icons instead of constructed. Moved "Play/Draw" winrates after ranked winrates.
image

lusbenjamin:
The history page will now display constructed ranked winrates for any time period selection (but not steps until next rank or limited ranked winrates).
image

lusbenjamin:
Okay, this is as far as it makes sense to go in this PR. Handling limited ranked stats more generally will require adding events from prior seasons to our data model somewhere.

PR #362, opened 2019-05-19
Check player id for each log message that gives it.

dougmill:
This should prevent direct challenges with both players on the same computer from corrupting gameplay data. Not sure how often that actually happens, but it's still good to have.

Also switched to regex literals.


gavacho:
Thanks for contributing @dougmill! These changes look very good

PR #364, opened 2019-05-19
opt-in git hooks

lusbenjamin:
### Motivation
- Open source projects should have a low barrier to entry.
- Opinionated is good, but opt-in convenience is better

Here are some git hooks that I have been using and enjoying:
- They are opt-in, so contributors may choose to use them or ignore them as they wish.
- They auto-install upon first run
- the pre-commit hook follows approach #5 listed here: https://prettier.io/docs/en/precommit.html
- the pre-push hook ensures that the build will pass before ever pushing a commit

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,77 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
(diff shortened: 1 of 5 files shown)


Manuel-777:
Nice ! 
I was not merging the other one because *I* did not like commits taking so long, but other contributor had issues getting prettier to work properly, so opt-in is perfect :)

PR #368, opened 2019-05-19
Deck details filters

lusbenjamin:
### Motivation
The deck details page uses completely unfiltered data! 🙀 

This PR fixes the situation, ensuring that the numbers and charts on the deck details page match the filters you had selected before getting there.

### Complications
- Adding a side panel to `ux_1` required finding all the places we used it and doing correct "resets" of CSS classes (similar design issue to de-registering event listeners)

### Bonus Good Stuff
- Removed deprecated `renderer.getDeckWinrate` math
- ~~Only show wildcards stats section if deck is incomplete~~
- Only show stats panel if there are matches in the currently selected filter history

### Before
image

### After
image


Manuel-777:
Having the actual wildcard cost of the deck could be helpful for sharing to others?

lusbenjamin:
> Having the actual wildcard cost of the deck could be helpful for sharing to others?
A good point that I didn't think about. I restored the rarity stats panel to always appear.

PR #375, opened 2019-05-20
Remove side effects from deck-drawer

gavacho:


gavacho:
Create an element and return it to callers instead of appending it to a specified element

PR #376, opened 2019-05-21
Remove globals defined in util.js

gavacho:
I want to get code out of util.js since all that code is global.


gavacho:
All of these commits remove functions from util.js. Some were never used anywhere. Some were used in only one other file -- these functions were moved into that file.

PR #382, opened 2019-05-22
Metadata Singleton

lusbenjamin:
### Motivation
This is a major refactor aimed at standardizing how the project consumes and updates card and collection metadata. The primary goals are:
  - centralize all logic that handles updating metadata
  - limit instantiation of metadata to one instance per process
  - remove as many global references as possible, especially mutations

### Approach
- Update `Database` class to explicitly handle all non-card metadata keys
- Update `Database` class to handle IPC updates
- Remove all global references to `cardsDb`, `setsList`, `eventsList`, `eventsToFormat`, `rankedEvents`.
- Remove old IPC `set_db` handlers in `renderer` and `overlay`
- Remove old mutations to globals in `background`
- ~~Create a new singleton factory `Database.getDb()`~~
- Convert `Database` module into a singleton
- Replace all former downstream consumers with local calls e.g. `db.card(grpId)`
  - modules can use `require` to directly import `Database`
  - non-modules may still rely on a `global` reference to `db` from `util` (but importing a new `Database` would return the same reference)


shared/__tests__/card-types-spec.js
```diff
@@ -24,9 +24,10 @@ describe("card-types", () => {
     });
 
     it("can determine the card type of any card except City's Blessing", () => {
-      Object.values(cardsDb.getAll()).forEach(card => {
+      cardsDb.cardList.forEach(card => {
         if (!_.has(card, "name")) return; // some properties are not cards :(
         if (card.name === "City's Blessing") return; // has no type
+        if (card.id === 100) return; // has invalid type
```
gavacho: This was necessary when the fake card was added to the permanent database. I believe that change has been reverted. If it has, then this line of code should be removed

lusbenjamin:
Based on a variety of great feedback from @AnnanFay, @Manuel-777, and @gavacho, the interface to the metadata singleton is now more concise:
- `db.card(grpId)`
- `db.sets[set].scryfall`
- `db.events_format[eventId]`
- `db.cardList.forEach(card =>...)`

lusbenjamin:
This now includes the season metadata as well, formerly handled manually as `season_start`, `season_end`, and some custom logic entangled with `updateRank`.

PR #387, opened 2019-05-24
Momir and shuffledOrder fixes.

dougmill:


window_background/gre-to-client-interpreter.js
```diff
@@ -549,10 +523,36 @@ GREMessages.GREMessageType_QueuedGameStateMessage = function(msg) {
...
+    currentMatch.turnInfo = {};
+    currentMatch.playerCardsUsed = [];
+    currentMatch.oppCardsUsed = [];
+    initialLibraryInstanceIds = [];
+    idChanges = {};
+    instanceToCardIdMap = {};
+  }
+  if (msg.msgId) {
+    currentMatch.msgId = msg.msgId;
+  }
```
dougmill: Apparently some messages (that have `msgId`) of other types can be logged out of order. I think order is reliable within a message type, though.

window_background/labels.js
```diff
@@ -289,7 +289,10 @@ function onLabelInEventGetPlayerCoursesV2(entry, json) {
 function onLabelInEventGetPlayerCourse(entry, json) {
   if (!json) return;
 
-  if (json.Id != "00000000-0000-0000-0000-000000000000") {
+  if (
+    json.Id !== "00000000-0000-0000-0000-000000000000" ||
+    json.InternalEventName === "Lore_WAR1_Momir"
+  ) {
```
dougmill: This event log is not properly filled out until you've completed a match, which resulted in my first match record of the event not having a deck.

dougmill:
Found another log that does have the deck before the match starts, so now the solution is general purpose. I'm still wondering if we'll have to do anything for the other 4 lore events, which all allow editing your deck during the event, though.

PR #404, opened 2019-05-31
Player Data Singleton

lusbenjamin:
### Motivation
This sprawling refactor has several goals:
- extract and centralize (most of) the global variables related to Arena player data
- change player data flow from [main pulls from background] to [background pushes to main]
  - enable auto-refresh when new player data arrives!
- untangle (most of) the IPC signals related to Arena player data
- move towards practice of treating player data model as immutable with one-way flow

### Approach
Since most of this data already lives together in the player config, the main approach was to keep `playerData`, `settings`, `cards`, `cardsNew`, `economy`, `history`, etc. all on the same object. Based on the prior success of #382, this PR also attempts to make that object into a singleton similar to the metadata `database`. This singleton has instances in each of the 3 renderer processes, background, main, and overlay. The IPC signals were reduced to only "set_player_data", and most of the complex filtering logic now lives in `background-util`.

### Global Data
- Reduces overall global variables across shared, background, main, and overlay code
- Creates 3 new global `pd` singletons, one in each process
  - Each singleton may be imported as a module and accessed similar to `db` metadata singleton
- Player data should now be handled as immutable. There is a bunch of Object and Array literals and having fun with spread syntax (some of this looks suspiciously like redux...).
- Updating data in the player-data singleton now all takes place in the `background` process and flows through `background-util.pd_set`.

### IPC Signals
- new "set_player_data" IPC signal
  - sent by `background` config reader, label parser, and remote data sync
  - received by `main` and `overlay` player-data singletons
- remove deprecated "set_settings", "set_cards", "set_decks", "set_economy", etc. IPC signals
- new "settings_updated" and "player_data_refresh" IPC signals
  - sent by `background` during `settings.loadSettings` and UI action handlers (tags, archiving)
  - received by `main` and `overlay` to handle old jQuery UI updating biz and page refreshes
- removed all "request_foo_data" signals that main used to send to background for each page load. Data changes are now ONLY pushed by `background` to the other player data singletons instead.

### Refactors
- `background.loadPlayerConfig` complete overhaul. Now just sticks the player store data directly into the player-data singleton
- tool launch has slightly better popup messages
- minor clean up of log statements during initial skim of log for arena ID
- grouped together economy, history, and event archival logic
- removed unused `decks_last_used`
- ~~`calculateRankWins` is only ever called when the user clicks one of their ranked icons~~
- refactored away `calculateRankWins`

### Misc Bugfixes
- Archiving a deck will no longer be overridden by restarting the tool while the deck is mentioned in the Arena log
- Expanding an event row now matches recent-first order in the rest of the tool
- Various edge cases during tool launch could cause errors that caused the taskbar progress to look like it was loading indefinitely. Most of these cases are now handled.
- Settings page slider behavior more consistent

### Other Noticeable Effects
- Launching the tools seems much faster for my typical use cases
- Page transitions seem much faster
- ~~The estimated number of games until next rank display takes a LOOOOOONG time to show up because reasons~~

### Potential Future Work
- ~~Archival event handlers and deck/match tag handlers still seem like a tangle~~
- ~~Sending data updates between singletons is still complex and easy to get wrong the first time~~
- ~~Calculating ranked winrates is horrendously slow (it was before, but now it's obvious)~~
- I noticed we have a tendency to slide towards circular dependencies. If we do another refactor to move things like `ipc_send` out of `renderer` into a new `main-util` module (similar to `background-util`), we can eliminate many of those circular dependencies from the main window code.
- Once the player data is all managed in one place, it should be much easier to refactor how we store it. Do we want to standardize how our data model is stored a little more? Index or no index? All docs directly on the main object or in nested objects like `decks`? Clean up any of those old keys? Separate denormalized `tool_deck_custom_data` to explicitly store tags/archived/other tool-only data?


shared/util.js
```diff
@@ -1,11 +1,5 @@
-/*
-global
-  cards
-  cardQuality
-  decks
-*/
```
lusbenjamin: In honor of @gavacho, I had to get the last globals out of `util`.

shared/player-data.js
```diff
@@ -0,0 +1,353 @@
...
+  courses_index: [],
+  matches_index: [],
+  draft_index: [],
+  gems_history: [],
+  gold_history: [],
+  decks: {},
+  staticDecks: new Set(),
+  decks_index: [],
+  decks_tags: {},
+  decks_last_used: [],
```
lusbenjamin: since I copied this constant without aggressively removing old keys, there is definitely cruft left in it. I remember removing `decks_last_used` from most of the rest of the code base.

shared/player-data.js
```diff
@@ -0,0 +1,353 @@
...
+}
+
+class PlayerData {
+  constructor() {
+    if (PlayerData.instance) return PlayerData.instance;
+
+    this.handleSetPlayerData = this.handleSetPlayerData.bind(this);
+    if (ipc) ipc.on("set_player_data", this.handleSetPlayerData);
+    this.handleSetCards = this.handleSetCards.bind(this);
+    if (ipc) ipc.on("set_cards", this.handleSetCards);
```
lusbenjamin: we may be able to continue refactoring to generalize and/or combine some of these remaining signals.

shared/player-data.js
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  handleSetDecks(_event, arg) {
+    this.staticDecks = new Set();
+    arg.forEach(deck => {
+      const deckData = {
+        // preserve custom fields if possible
+        ...(this.deck(deck.id) || {}),
+        ...deck
+      };
```
lusbenjamin: ^ this style of object assignment is how the PR removes "hexproof from archiving"

window_main/aggregator.js
```diff
@@ -77,29 +66,6 @@ class Aggregator {
...
-      const match = matchesHistory[mid];
-      if (!match || match.type === "draft" || match.type === "Event") return;
-      try {
-        if (match.playerDeck && match.playerDeck.mainDeck) {
-          match.playerDeck.colors = get_deck_colors(match.playerDeck);
-        } else {
-          match.playerDeck = JSON.parse(
-            '{"deckTileId":' +
-              DEFAULT_TILE +
-              ',"description":null,"format":"Standard","colors":[],"id":"00000000-0000-0000-0000-000000000000","isValid":false,"lastUpdated":"2018-05-31T00:06:29.7456958","lockedForEdit":false,"lockedForUse":false,"mainDeck":[],"name":"Undefined","resourceId":"00000000-0000-0000-0000-000000000000","sideboard":[]}'
```
lusbenjamin: just in case this data normalization still matters, I moved it into the new player data singleton. if we want to remove it later, hopefully the data migrations will be easier to handle there

window_main/decks.js
```diff
@@ -53,225 +49,216 @@ function setFilters(selected = {}) {
 
 //
 function openDecksTab(_filters = {}) {
-  if (sidebarActive == 0 && decks != null) {
```
lusbenjamin: most of the changes in this file are because I removed this indentation block. mea culpa, should have left it alone to make the refactor more clear.

window_main/decks.js
```diff
@@ -374,38 +361,34 @@ function createTag(tag, div, showClose = true) {
...
+  if (!deck || deck.format === tag) return;
+  if (deck.tags && deck.tags.includes(tag)) return;
 
-  let obj = { deck: deckid, name: tag };
-  ipc_send("add_tag", obj);
+  const decks_tags = {
+    ...pd.decks_tags,
+    [deckid]: [...deck.tags, tag]
+  };
+  pd.handleSetPlayerData(null, { decks_tags });
```
lusbenjamin: this logic is now identical to the logic in `background` that handles "add_tag". which is probably a great sign! but also an opportunity to potentially refactor, maybe move into `player-data` some how?

window_main/renderer.js
```diff
@@ -142,6 +126,12 @@ function ipc_send(method, arg, to = IPC_BACKGROUND) {
   ipc.send("ipc_switch", method, IPC_MAIN, arg, to);
 }
 
+// convenience handler for player data singleton
+function toggleArchived(id) {
+  pd.toggleArchived(id);
+  ipc_send("toggle_archived", id);
```
lusbenjamin: even though the player data all lives happily together in the `pd` singleton, it's actually a tringleton: one `pd` in background, main, and overlay each. The code above follows a pretty common pattern: update the local singleton, tell IPC to get the other 2 singletons to catch up if necessary.

window_main/renderer.js
```diff
@@ -264,33 +249,16 @@ ipc.on("set_player_data", (event, _data) => {
 });
 
 //
-ipc.on("set_decks_last_used", (event, arg) => {
-  playerData.decks_last_used = arg;
+ipc.on("rank_wins_updated", () => {
+  openHistoryTab();
```
lusbenjamin: this and "calculate_rank_wins" are some of the last vestiges of the old sync-style request-upon-load architecture. Most of the rest of the pages are already g2g

window_background/background.js
```diff
@@ -425,100 +328,45 @@ ipc.on("delete_data", function() {
...
-ipc.on("unarchive_economy", function(event, _id) {
-  ipc_send("show_loading");
-  economy[_id].archived = false;
-  store.set(_id, economy[_id]);
+  ipc_send("popup", {
+    text: "Calculating ranked win rates...",
+    time: 0,
+    progress: 2
+  });
+  calculateRankWins();
```
lusbenjamin: this is where I chose to hide the most expensive remaining part of page transitions. it only gets called when you click on one of the ranked icons. it's painfully slow to wait for this right now.

window_background/background.js
```diff
@@ -695,236 +554,129 @@ ipc.on("set_odds_samplesize", function(event, state) {
...
+  data.courses
+    .filter(doc => !pd.eventExists(doc._id))
+    .forEach(doc => {
+      const id = doc._id;
+      doc.id = id;
+      delete doc._id;
+      courses_index.push(id);
+      store.set(id, doc);
+      pd_sync({ [id]: doc });
+    });
```
lusbenjamin: I have not yet tested any of the patreon user data sync code. I think this is it?

shared/player-data.js
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  handleSetPlayerData(_event, arg) {
+    const data = { ...arg };
+    delete data.changes;
+    delete data.drafts;
+    delete data.events;
+    delete data.matches;
+    Object.assign(this, data);
+  }
```
gavacho:
```
  handleSetPlayerData(_event, arg) {
    const bannedKeys = ["changes", "drafts", "events", "matches"]
    Object.assign(this, _.omit(args, bannedKeys));
  }
```

shared/util.js
gavacho: i am honored

window_background/background.js
lusbenjamin: Seems to work well for my new vintage status! 🌟 🎩 🌟 After, @Manuel-777 enabled it for my account, I received several previously-deleted old matches successfully.

lusbenjamin:
Refactored out the super slow `calculateRankWins` and tested the remote data sync.

shared/player-data.js
lusbenjamin: sexy! I need to learn some more `lodash` 😸

window_background/background.js
```diff
@@ -273,13 +272,11 @@ ipc.on("login", function(event, arg) {
   if (arg.password == HIDDEN_PW) {
     tokenAuth = rstore.get("token");
     httpApi.httpAuth(arg.username, arg.password);
-    httpApi.httpHomeGet();
   } else if (arg.username === "" && arg.password === "") {
     offlineLogin();
   } else {
     tokenAuth = "";
     httpApi.httpAuth(arg.username, arg.password);
-    httpApi.httpHomeGet();
```
Manuel-777: Removed the initial loading from the home tab, this was causing (for me at least) a premature IPC `"set_offline"` before login, thus causing renderer process to think I was offline constantly.

window_background/http-api.js
```diff
@@ -106,7 +106,7 @@ function httpBasic() {
           protocol: "https:",
           port: 443,
           hostname: serverAddress,
-          path: "/database/database.json",
+          path: "/database/",
```
Manuel-777: new database endpoint is also smarter! I can rollback changes and push updates to the same endpoint, and individual or previous database files they will be accessible using their version number; `https://mtgatool.com/database/1.json`

window_background/background.js
lusbenjamin: That makes sense and is a much better idea! Differences in our network topology explains why it only worked for me.

shared/player-data.js
```diff
@@ -0,0 +1,303 @@
...
+}
+
+class PlayerData {
+  constructor() {
+    if (PlayerData.instance) return PlayerData.instance;
+
+    this.handleSetData = this.handleSetData.bind(this);
+    if (ipc) ipc.on("set_player_data", this.handleSetData);
+    this.handleMergeData = this.handleMergeData.bind(this);
+    if (ipc) ipc.on("merge_player_data", this.handleMergeData);
```
lusbenjamin: The new player data IPC signals have been reduced to just "set_player_data" for destructive changes and "merge_player_data" for additive ones.

shared/player-data.js
```diff
@@ -0,0 +1,303 @@
...
+
+    PlayerData.instance = this;
+  }
+
+  handleSetData(_event, arg) {
+    Object.assign(this, arg);
+  }
+
+  handleMergeData(_event, arg) {
+    _.merge(this, arg);
```
lusbenjamin: All of the power behind the new "merge_player_data" additive updates comes from [`lodash.merge`](https://lodash.com/docs/4.17.11#merge)

window_background/background-util.js
```diff
@@ -61,8 +68,59 @@ function unleakString(s) {
...
+function pd_merge(data) {
+  const cleanData = _.omit(data, dataBlacklist);
+  pd.handleMergeData(null, cleanData);
+  pd_sync(cleanData, "merge_player_data");
+}
+
+function pd_sync(cleanData, signal) {
+  ipc_send(signal, cleanData, IPC_MAIN);
+  const overlayData = _.pick(cleanData, overlayWhitelist);
+  ipc_send(signal, overlayData, IPC_OVERLAY);
```
lusbenjamin: we now only send the necessary subset of player data to the overlay singleton (should be the same data as `master` uses)

lusbenjamin:
@Manuel-777 this PR now finally hit the Holy Grail of auto-refresh! 🎉 

The only remaining problem is attempting to remember scroll bar position / last data index. Can you work your overlay auto scroll magic on the main window pages?

shared/player-data.js
lusbenjamin: RIP `lodash.merge`. Removed in favor of keeping updates to IPC and `electron-store` more consistent and refactoring down to a single Elder Dragon Signal.

lusbenjamin:
Auto-refreshes now remember the last data index and scroll position of the user. Archival toggle animations are slightly less awkward. I would be happy shipping this at this point.

window_main/list-item.js
```diff
@@ -44,8 +44,7 @@ class ListItem {
       this.deleteButton.addEventListener("click", e => {
         e.stopPropagation();
         this.onDeleteCallback(this.id);
-        this.container.style.height = "0px";
-        this.container.style.overflow = "hidden";
```
Manuel-777: Even trough changes must be submitted by background for aprooval I think we should still update the UI elements upon interaction, for a more responsive UX, We would only *not* update the database in renderer, but waiting for background's signal and imediate refresh makes for a poor interaction of something simple as hiding an item.

lusbenjamin:
I always forget about Tournaments the first time. And the second. Fortunately today was the third time, so I bugfixed the deck selector on the join tournament page. While I was at it, I made it a little smarter and shinier:
image

PR #407, opened 2019-06-05
No Globals in Main Window Process

lusbenjamin:
### Motivation
- No more globals in any of the main window process files.
- Eliminate circular dependencies in `require` chain

### Approach
- Split `renderer` into 2 files:
  - `renderer-util` for fns shared across files in main window process
  - `renderer` Classic for IPC listeners, navigation logic, and "outer" window chrome
- Circular `require` dependencies should now be linear chains, e.g. `renderer` -> `decks` -> `deck-details` -> `renderer-util`
- Split up global variables and refactor into metadata database, player data singleton, or `renderer-util.localState` depending on usage

### DIFF
--- a/shared/database.js
+++ b/shared/database.js
@@ -6,8 +6,10 @@ class Database {
   constructor() {
     if (Database.instance) return Database.instance;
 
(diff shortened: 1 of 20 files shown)


lusbenjamin:
image

PR #413, opened 2019-06-09
Multiple overlays

Manuel-777:
Upgrade overlay to level 1000!

- Settings revamped to allow up to 4 overlay windows at the same time.
- All overlays behave independently from each other.
- Clear separation between normal overlay and draft overlay.
- Reworked some of the flow of data between settings saving and the overlays.

Still testing to be done, especially with draft *(make a draft run, see if the draft overlay pops correctly, finish, it should hide, then start a match, the match overlay should pop without showing the draft overlay anymore)*, but most normal use cases seem to behave properly.

Demo image:
image


shared/player-data.js
```diff
@@ -42,37 +46,90 @@ const playerDataDefault = {
...
+        lands: true,
+        mode: 1,
+        ontop: true,
+        scale: 100,
+        show: true,
+        show_always: false,
+        sideboard: false,
+        title: true,
+        top: true
+      },
```
lusbenjamin: you could make a template `overlaySettingsTemplate` and then use `objectClone` to make this more like:
```
const overlaySettingsTemplate = {
  alpha: 1,
  alpha_back: 1,
  bounds: { width: 300, height: 600, x: 0, y: 0 },
  ...,
  title: true,
  top: true
};

const defaultCfg = {
  ...,
  settings: {
    sound_priority: false,
    sound_priority_volume: 1,
    ...,
    last_open_tab: -1,
    card_tile_style: CARD_TILE_FLAT,
    overlays: [
      {
        ...objectClone(overlaySettingsTemplate),
        mode: 1
      },
      {
        ...objectClone(overlaySettingsTemplate),
        mode: 2
      },
```

window_main/renderer.js
```diff
@@ -537,7 +544,7 @@ function force_open_settings() {
     }
   });
   $(".moving_ux").animate({ left: "0px" }, 250, "easeInOutCubic");
-  openSettingsTab();
+  openSettingsTab(section);
```
lusbenjamin: should probably be `openSettingsTab(section, 0)`

window_overlay/overlay.js
```diff
@@ -999,47 +994,30 @@ $(document).ready(function() {
...
-    deckMode += 1;
-    if (deckMode > 4) {
-      deckMode = 0;
-    }
-    updateView();
-  });
-
   //
   $(".close").click(function() {
     ipc_send("overlay_close", 1);
```
lusbenjamin: this signal needs to be updated to use `overlayIndex`

window_main/settings.js
```diff
@@ -652,61 +788,68 @@ function updateUserSettingsBlend(_settings = {}) {
...
-    overlay_clock: overlayClock,
-    overlay_sideboard: overlaySideboard,
-    overlay_ontop: overlayOnTop,
-    overlay_lands: overlayLands,
     anon_explore: anonExplore,
     back_color: backColor,
     back_url: backUrl,
     export_format: exportFormat,
     skip_firstpass: !readonlogin,
+    ...pd.settings,
```
lusbenjamin: this line must change to:
`overlays: [...pd.settings.overlays],`

lusbenjamin: ❗️ ❗️ ❗️ To be clear, as written this PR will _only_ correctly handle user updates to overlay settings. It currently ignores updates to other settings, e.g. changing the background image.

main.js
```diff
@@ -565,27 +603,91 @@ function createOverlay() {
...
+  over.on("move", function() {
+    //console.log(new Date().getTime());
+    let timer = overlaysTimeout[index];
+    if (timer) {
+      clearTimeout(timer);
+    }
+    overlaysTimeout[index] = setTimeout(function() {
+      saveOverlayPos(index);
+      overlaysTimeout[index] = null;
+    }, 100);
```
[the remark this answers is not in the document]
Manuel-777: Yes, resize does not update when we only move the window, and using the callback without the timeout like this spammed our IPC like hell

shared/player-data.js
Manuel-777: I had it set like that initially but I wanted to set default settings on each overlay, so users could see the different options combinations already set up

main.js
```diff
@@ -543,49 +562,122 @@ function createMainWindow() {
...
+
+  over.once("did-finish-load", function() {
+    over.webContents.send("set_overlay_index", index);
+  });
+
+  return over;
+}
+
+function saveOverlayPos(index) {
+  if (!overlays[index] || overlaysResizeLock) return false;
```
Manuel-777: This was an interesting one. Basically what happened was that, using persistent overlay (which I do a lot during debug), the overlay started at default position until user settings were loaded, thus causing them to tell background they moved into a new position right after loading. User settings triggered the overlays to move into a new (previously saved) position, thus triggering a window move event. Telling the background they moved again. This made all overlays with non-default positions and presistent mode active go crazy and start flipping all over the screen.
It was PROBABLY caused in the first place because I was cleaning up the position saving code for the overlays, and with the delays from having multiple IPC calls we did not notice. Now the flow is more direct between main and background it was quite a show of flashing windows :)

lusbenjamin: Totally wild! I will do some more testing with persistent overlay during startup today. I am starting to feel like we may have to address performance immediately after merging this PR.

window_background/background.js
```diff
@@ -292,8 +292,7 @@ ipc.on("windowBounds", function(event, obj) {
 //
 ipc.on("overlayBounds", function(event, index, obj) {
   pd.settings.overlays[index].bounds = obj;
-  pd_set({ settings: pd.settings });
-  store.set("settings", pd.settings);
```
lusbenjamin: with the call to `store.set` removed, how does the new position ever get saved? Does the user have to change some other setting as well to see the position change persist?

PR #416, opened 2019-06-12
Restore Tag Autocomplete and Misc Tag QoL

lusbenjamin:
This is a small collection of UX quality-of-life upgrades related to tags.
- Restores autocomplete functionality on the history page. Now combines your existing tags and the metadata archetypes into a unified list.
- ~~Uses lower case for all tags and archetypes on Decks and History page~~
- ~~Uses case-insensitive logic whenever handling tags~~
- Treats all loss-of-focus events as save requests, i.e. like hitting "Enter"
- Better logic to handle some input edge cases


lusbenjamin:
We lost autocomplete functionality as a side effect of the work we did in in these commits:
- https://github.com/Manuel-777/MTG-Arena-Tool/commit/5c4b88fc6e625cb9538c41953fc9c5f8ab7f9a26
- https://github.com/Manuel-777/MTG-Arena-Tool/commit/2a226d421f14c0202b050c6450c041b81e97b2b3

lusbenjamin:
Since the history page uses `autocomplete`, it was having conflicts between the blur handler I added and the one that had unwittingly re-enabled from `autocomplete`. Although using the dropdown with the arrow keys and Enter worked fine, clicking an item in the dropdown or anywhere else in the document did not reliably produce the desired submissions. Moving all of the responsibility to a single blur handler in `autocomplete` solved the problem!

Issue #282, opened 2019-04-25
Remember Draft Pack-Pick History for Non-Continuous Drafts


lusbenjamin:
I believe the current tool behavior only saves draft pack-pick options and player selections at the very end of a draft. If the MTG Arena Tool is not running continuously for the entire drafting session, pack-pick history from earlier sessions may be lost entirely.

![image](https://user-images.githubusercontent.com/14894693/56710468-2704ae80-66db-11e9-8faa-5e85e4072abc.png)

PR #434, opened 2019-06-22
Bugfix Draft Data Persistence

lusbenjamin:
### Motivation
This PR ensures that we save incremental draft data, which allows users to stop and restart drafts without losing all previous pick/pack info.
https://github.com/Manuel-777/MTG-Arena-Tool/issues/282

### Other Bugfixes
- Ensures that match and draft overlays close whenever the player navigates to Arena Home screen
- Final draft object correctly shows final pick (had no effect on draft replays, but bothered me anyway)
- Draft overlays are more robust to transparency changes, no more amnesia if they are recreated
- Draft overlays now guaranteed to have latest collection card-count data
- Draft overlays now correctly stay hidden during first log pass

### Refactor Notes
- Removes background global `currentDraft` since we now read and write from the store
- groups together draft related fns in `background` - these handle IPC signals and store access
- groups together draft-related fns in `labels` - these translate data structure
- logic to start a draft no longer resets match state (I think this is okay?)
- preserve original `draftId` throughout: this is what we now use to tie together log entries across time
  - there is now one additional non-standard draft object in `player-data` that acts as swap space, this exactly mirrors the json data we have in the logs
- preserve as much of the original log data as we can with fancy spread operator

### DIFF
--- a/window_background/background.js
+++ b/window_background/background.js
@@ -1644,7 +1644,7 @@ function setDraftData(data) {
 //
 function endDraft(data) {
   duringDraft = false;
-  if (!firstPass) ipc_send("set_arena_state", ARENA_MODE_IDLE);
+  if (debugLog || !firstPass) ipc_send("set_arena_state", ARENA_MODE_IDLE);
   httpApi.httpSetDraft(data);
   ipc_send("popup", { text: "Draft saved!", time: 3000 });
 }
--- a/window_background/background.js
+++ b/window_background/background.js
@@ -39,7 +39,8 @@ const {
   HIDDEN_PW,
   IPC_OVERLAY,
   ARENA_MODE_MATCH,
-  ARENA_MODE_DRAFT
+  ARENA_MODE_DRAFT,
+  ARENA_MODE_IDLE
 } = require("../shared/constants");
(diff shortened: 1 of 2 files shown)


Manuel-777:
`Draft overlays now correctly stay hidden during first log pass`
Are we using `if (debugLog)` in this case? I was using that to debug drafts in the past.

`logic to start a draft no longer resets match state (I think this is okay?)`
Yes, matches and drafts should not mess up with eachother, this was one of the goals of the multiball project as well, so its great!

window_background/background.js
```diff
@@ -1626,26 +1604,33 @@ function saveMatch(id) {
...
+  return _.cloneDeep(currentDraftDefault);
+}
+
+//
+function startDraft() {
+  if (debugLog || !firstPass) {
+    if (pd.settings.close_on_match) {
+      ipc_send("renderer_hide", 1);
+    }
+    ipc_send("set_arena_state", ARENA_MODE_DRAFT);
```
lusbenjamin: `startDraft` should respect both `debugLog` and `firstPass` correctly, though I did not test `debugLog` yet. There is a new `duringDraft` local state variable that we now use like `duringMatch`.

PR #437, opened 2019-06-24
Enhance First Pass

lusbenjamin:
### Motivation
This PR optimizes several parts of the initial tool launch. It has 2 major goals:
- For users who always read the entire log, make it ~~blazingly fast~~ a little faster
- For users with `skip_firstpass`, make it more robust upon launch

### Approach
- Add default values and move several fields into `player-data` storage: e.g. `rank`, `economy`, `static_decks`, `newCards`, etc.
  - Since we load all `player-data` in bulk, this automatically makes `skip_firstpass` much more robust (e.g. displays wildcards, decks, and ranks immediately)
  - Adds `store.set` calls to keep these fields up-to-date like other fields. This would make reading the log all at once even slower except...
- During `firstpass`, skip all IO calls to `store` except one giant `store.get` at the beginning and one giant `store.set` at the end
- Also add a better description of the `skip_firstpass` feature to the settings page:
image

### DIFF
--- a/shared/database.js
+++ b/shared/database.js
@@ -17,6 +17,7 @@ class Database {
 
     this.rewards_daily_ends = new Date();
     this.rewards_weekly_ends = new Date();
+    this.activeEvents = [];
 
     const dbUri = `${__dirname}/../resources/database.json`;
     const defaultDb = fs.readFileSync(dbUri, "utf8");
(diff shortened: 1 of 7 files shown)


Manuel-777:
One idea I had some time ago was to analyze if a match had already been processed and if so ignore it when reading the log (most GRE messages have the match ID) so we can identify those messages and simply not analyze them if they correspond to a match id already in the player config. It might be a little speed up a little bit (not drastically) but every bit counts :)

lusbenjamin:
forgot to mention that this PR also completely removes the deprecated `decks_index`

PR #448, opened 2019-06-28
centralize player_data_refresh IPC

lusbenjamin:
### Motivation
This is a "relatively safe" refactor focused on centralizing our `"player_data_refresh"` signals.

### Approach
- `background-util.setData` now takes explicit `refresh` param which sends `"player_data_refresh"`
  - smart default to `debugLog || !firstpass`
- remove (now redundant) old calls to `ipc_send`
- in places where we called `setData` _without_ an old call to `ipc_send`, prevent refresh signal if necessary

### Bonus Refactors
- combined `"player_data_updated"` and `"player_data_refresh"`
- `syncSettings` also accepts a `refresh` param
- grouped together some adjacent calls to `setData` to minimize signals
- removed last traces of deprecated `decks_index` from code
- made sure some other `firstPass` logic respected `debugLog`

PR #454, opened 2019-06-29
bugfix overlays launch properly when tool starts mid match

lusbenjamin:
This fixes the logic for showing overlays when the tool is launched mid-Arena-match
(I may have broken it during #448)


lusbenjamin:
_"relatively safe refactor"_ :roll_eyes:

Issue #443, opened 2019-06-26
Need to click "continue offline" on every start


Porkepix:
If a user want to use the application without account, for just local features on the computer, he needs to click "continue offline" as you get the Welcome screen on every single start.

It should be possible to set definitely that you want to use it without account.

lusbenjamin:
This feature request is my white whale. I have tried 3 times to get this to work, and each time I have failed for different reasons. 🤕 

Maybe the 4th time will be the charm!

PR #464, opened 2019-07-04
Offline auto-login

lusbenjamin:
### Motivation
This is a grudge match. I attempted to get auto-login to work as one of my first contributions to the project, and ever since I have been struggling to get this feature to work for users in offline mode. A few different users have requested this, most recently in https://github.com/Manuel-777/MTG-Arena-Tool/issues/443.

### Approach
The delicate tangle of bootstraps and spaghetti code at the launch of our app continues to baffle me, but here is my latest idea:
- Change the first time we let users access the login page / initiate the auto-login feature:
  - used to be when we receive a response for the initial metadata DB request (or timeout)
  - now when we complete a successful initial pull of `arenaID` and `name` from the fast `logLoop`

### Bonuses
- Update settings help text to make sharing settings and auto-login clearer
- Update offline splash page to give more practical suggestions in a variety of cases
- bugfix: do not spam the user with UI refreshes when a bunch of queries fail at once
- bugfix: ensure nav bar icons are always in the correct sized mode for certain edge cases

### DIFF
--- a/main.js
+++ b/main.js
@@ -35,7 +35,6 @@ var mainTimeout = null;
 
 var tray = null;
 var closeToTray = true;
-let autoLogin = false;
 let launchToTray = false;
 
 const ipc = electron.ipcMain;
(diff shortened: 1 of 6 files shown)


lusbenjamin:
### New Offline Splash
image

### New

PR #471, opened 2019-07-11
Bugfix mastery track

lusbenjamin:
### Motivation
Wizards gonna Wiz.

> "it seems highly likely to me that we will see the Arena mastery track data model change one more time when they release the next one"

### Dirty Details
- the log contains "inventory change" entries and "track progress" entries
  - both types of entry have lots of redundant data (e.g. gold delta, gem delta, exp delta)
  - each type of entry has some unique data (e.g. orb delta only on "track progress", `aetherizedCards` only on "inventory change"
- since we really want to show all kinds of data to the user, we have to blend the two together
  - original build (before weekly) showed experience delta on "track progress"
  - this PR shows experience delta on "inventory change" and hides all non-level-change "track progress" entries
- bonus bugfix: `labels.onLabelTrackProgressUpdated` correctly uses log entry timestamp
- bonus refactor: Economy page no longer relies on any modulo arithmetic for the vault or exp delta
- bonus bugfix: standardizes all percentage formatting on Economy page
- bonus enhancement: "track progress" context more specific
- bonus enhancement: handles mastery pass purchase transactions

### Before
image

### After
image

### DIFF
--- a/window_background/background.js
+++ b/window_background/background.js
@@ -77,7 +77,8 @@ const {
   onLabelInEventGetSeasonAndRankDetail,
   onLabelGetPlayerInventoryGetRewardSchedule,
   onLabelRankUpdated,
-  onLabelTrackProgressUpdated
+  onLabelTrackProgressUpdated,
+  onLabelTrackRewardTierUpdated
 } = require("./labels");
 
 const toolVersion = electron.remote.app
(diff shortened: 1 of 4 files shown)


lusbenjamin:
Adding some improvements to "Track Progress" context based on feedback from @AnnanFay.
image

lusbenjamin:
Now handles Mastery Pass purchase:
image

PR #475, opened 2019-07-14
Overlays v3

Manuel-777:
### Motivation
Currently in "multiball" world we create **5** different overlay windows, one for each overlay. This had a bunch of downsides in practice that were quickly noticeable;
- Each overlay created a new renderer process that had to be constantly on in order to work properly, adding memory usage.
- Each overlay had to manage its own settings and data, and sometimes this meant sending 5 IPC messages, one for each window, even if they were not used.
- Whenever an overlay changed mode or new data was added there was a considerable delay for updates, that made it seem laggy and unresponsive at times.

We solve this creating a *single overlay instance* that lives on top of the screen constantly, in this screen we create all overlay elements as absolute divs that can be arranged and moved freely.
To avoid drawing to the screen even while arena is not running we check if the Arena process is focused (or if we want to see the overlays, such as when we want to edit the positions), and then we move the overlay window to cover the screen and draw the elements.

Overall the performance was vastly improved, along with the responsiveness of the overlay screen. we update on each change to al overlays and all overlays have perfect data of all events going on regardless of the mode we are drawing them.

image

The overlay screen is always click-through, this causes some issues when we actually want to make clickable elements (such as close buttons on navigation bars), to achieve this we make the entire overlay clickable only for while the element is hovered using `setIgnoreFalse()` and `setIgnoreTrue()` to revert it when its not hovered. We also use these functions to enable edit mode (`Alt + E`), in that case we ignore all buttons until edit mode ends to avoid locking the elements inside edit mode without being clickable.

### To do:
- [x] Fix clocks; Now each overlay requires a clock mode variable living inside overlay.js.
- [x] Test drafts.
- [x] global setting for scaling (and moved tweeter setting)
- [x] fixed "reset position" button in settings
- [x] Give more use to the tracking Tool logo on the game (turn edit mode on/off, more?)
- [ ] Optionally figure out some way to allow users to edit positions right from the settings panel. (requires forcing an overlay to be always on temporarily? requires some sort of confirm button other than the icon?)
- [ ] Add an option to select in which screen the overlay will be drawn


Rikahe:
This breaks the Tracker under Linux.

This [commit](https://github.com/Rikahe/MTG-Arena-Tool/commit/3a731178fdc6ce26e3d15bdb0197115d736c9428) atleast shows the overlays while any WINE window is active.

Never really did any JS dev so i have some problems navigating the codebase and i am not able to get dev-tools working for the overlay windows

- in checkActiveWindow(): editMode always evaluates to false *shrug so edit mode is not changing visibility
- the overlay is still not interact able
- i am playing the game windowed & capture the mouse (overlays on another monitor), so most of the active window checks do not make any sense as i have to make WINE/MTGA inactive to move the mouse to the overlays anyway

This is as far as i tried the only tracker that i could build/run without problems under Linux. Would be nice if this is still the case tomorrow :D

Manuel-777:
> This breaks the Tracker under Linux.
> 
> This [commit](https://github.com/Rikahe/MTG-Arena-Tool/commit/3a731178fdc6ce26e3d15bdb0197115d736c9428) atleast shows the overlays while any WINE window is active.
> 
> Never really did any JS dev so i have some problems navigating the codebase and i am not able to get dev-tools working for the overlay windows
> 
> * in checkActiveWindow(): editMode always evaluates to false *shrug so edit mode is not changing visibility
> * the overlay is still not interact able
> * i am playing the game windowed & capture the mouse (overlays on another monitor), so most of the active window checks do not make any sense as i have to make WINE/MTGA inactive to move the mouse to the overlays anyway
> 
> This is as far as i tried the only tracker that i could build/run without problems under Linux. Would be nice if this is still the case tomorrow :D

Triying to run the new modules on Arch without sucess to test, what are your specs? (distro, npm version, node version, etc)

Rikahe:
```
> node --version
v10.16.0

> npm --version
6.9.0

> lsb_release -a
No LSB modules are available.
Distributor ID: neon
Description:    KDE neon User Edition 5.16
Release:        18.04
Codename:       bionic
```

mtga with lutris install script running in virtual desktop mode

lusbenjamin:
https://interactjs.io/

lusbenjamin:
Re: @Rikahe running on Linux, I noticed that this new overlay sets `focusable = false`. From the electronjs documentation:
"focusable Boolean (optional) - Whether the window can be focused. Default is true. On Windows setting focusable: false also implies setting skipTaskbar: true. **On Linux setting focusable: false makes the window stop interacting with wm, so the window will always stay on top in all workspaces.**"
https://electronjs.org/docs/api/browser-window

lusbenjamin:
Also:
"On Linux, users have to put --enable-transparent-visuals --disable-gpu in the command line to disable GPU and allow ARGB to make transparent window, this is caused by an upstream bug that alpha channel doesn't work on some NVidia drivers on Linux."
https://electronjs.org/docs/api/frameless-window#transparent-window

lusbenjamin:
@Manuel-777 Looks awesome from the bit I've seen so far 😻! I made some quick changes:
- Replaced `jquery` and `jquery-ui` dependencies with `interactjs` (reduced the size of diff by an order of magnitude, also has some cool support for snap-to-grid or edge of window stuff if you want to take a look)
- Reenabled `jest` tests and fixed the Travis build

Manuel-777:
> @Manuel-777 Looks awesome from the bit I've seen so far ! I made some quick changes:
> 
> * Replaced `jquery` and `jquery-ui` dependencies with `interactjs` https://interactjs.io/ (reduced the size of diff by an order of magnitude, also has some cool support for snap-to-grid or edge of window stuff if you want to take a look)
> * Reenabled `jest` tests and fixed the Travis build

The travis thing was causing errors with ´active-win´ library I was attempting to use, I removed it because it wasnt capturing anything in my linux system, while on debian it worked perfectly. and I want the experience to be the same across all systems.

lusbenjamin:
An excellent concern about potentially stale data! I think we should be okay here because there is no way for data to get stale in the Uniball world, so we only have to worry about stale display. Since hidden overlay divs _should_ always attempt to re-render their contents when they are shown, they should always re-render the latest data.

window_overlay_v3/overlay.js
```diff
@@ -0,0 +1,1252 @@
...
+  }
+  actionLog.sort(compare_logs);
+  //console.log(arg.seat, arg.str);
+});
+
+ipc.on("settings_updated", settingsUpdated);
+
+function settingsUpdated() {
+  webFrame.setZoomFactor(pd.settings.overlay_scale / 100);
+  if (editMode) return;
```
lusbenjamin: Is there a specific reason that we need to dis-allow updating settings during edit mode? If so, I'd love a comment here to help me remember why.

Manuel-777: Yep, so that we can edit during play mostly. Else the window will keep being reset to its previous position while you hopelessly try to move it around.

Manuel-777:
> An excellent concern about potentially stale data! I think we should be okay here because there is no way for data to get stale in the Uniball world, so we only have to worry about stale display. Since hidden overlay divs _should_ always attempt to re-render their contents when they are shown, they should always re-render the latest data.

Just tested and indeed overlay windows "wait" for the next message to update stale data, I don't think we should be worrying so much about hidden divs drawing? at least in video it does not affect at all if its hidden, just during the processing.
Else we should force-populate the overlays when they change from hidden to visible, without using IPC, so its instantaneously shown.

window_overlay_v3/overlay.js
```diff
@@ -208,16 +206,19 @@ function saveOverlaysPosition() {
...
+    const overlayDiv = byId("overlay_" + (index + 1));
+    const forceInt = num => Math.round(parseFloat(num));
+    const dx = parseFloat(overlayDiv.getAttribute("data-x")) || 0;
+    const dy = parseFloat(overlayDiv.getAttribute("data-y")) || 0;
+    const bounds = {
+      width: forceInt(overlayDiv.style.width),
+      height: forceInt(overlayDiv.style.height),
+      x: forceInt(parseFloat(overlayDiv.style.left) + dx),
+      y: forceInt(parseFloat(overlayDiv.style.top) + dy)
+    };
```
Manuel-777: Im not sure how to fix this (I dont follow the logic on interactJs) , but pressing `Alt + Shift + E` repeatedly makes the overlays move in crazy ways after this commit? I guess we are doing something with saving the position that ends up changing the initial value but I cant be sure where it happens.

lusbenjamin: I am also just learning interactjs, so I started by copying and pasting their examples together. For some reason, they don't seem to play well with our current setup (maybe because we are using `position: abolute !important`?). I just pushed a commit that attempts to strip down the interactjs call to the absolute minimum.

window_overlay_v3/overlay.js
lusbenjamin: That makes sense! We probably want to add a small opacity/visibility to the master overlay wrapper during edit mode to make this less confusing and more obviously a modal-style interaction.

window_overlay_v3/overlay.js
Manuel-777: I think that did the trick, whatever it was windows seem to behave properly now :)

lusbenjamin:
I use the sublime editor and frequently change between 1- and 2-column layouts while coding. While developing on this PR, my Sublime editor suddenly stopped letting me change layouts. 🙀 

...then I realized that I was mashing "Shift+Alt+1" and "Shift+Alt+2" trying to change layouts and it was just that MTG Arena Tool now clobbers my dev shortcuts. 😿

Manuel-777:
We definitely needs a custom hotkey setter

PR #478, opened 2019-07-15
bugfix stale UI elements based on all matches aggregation

lusbenjamin:
### Motivation
Before we fully converted to the current push-from-background data flow, we had to maintain an "all matches" aggregation singleton to produce some UI elements. We have kept it (mostly) working, but the old approach does cause those UI elements to slowly "go stale" until the user completely restarts mtgatool.

### Approach
This PR cleans up the old dependencies on the aggregation singleton and replaces them with more specific and up-to-date aggregations (also faster in most cases).

### Bonus Bugfix
- handles when the user deletes the archetype from the last match with that archetype (before it would show an empty screen)


Manuel-777:
Merged before I could ask you to resolve! :nerd_face: 
Is it g2g now?

lusbenjamin:
g2g! 🚀

Issue #462, opened 2019-07-03
Improve setting non-standard log files locations with graphical selection


Porkepix:
Recent commits allowed onto a Linux install with unusual log files location to set it with a textbox where user have to write or paste the full path into the box. Besides having this way, allow user for a graphical selection would also be nice in those situations.

PR #497, opened 2019-07-22
Arena Data Settings Tab

lusbenjamin:
### Motivation
- Make it easier for @Manuel-777 to debug custom `output_log.txt` files
- Expose more of how mtgatool talks to Arena and help users self-diagnose
- Support international and non-Windows users with hard-to-predict environments

### Approach
Build a new settings page! WARNING: this also touches the logic in `backgound-util.parseWotcTime`, which has historically been dangerous code to mess with.
image

### Fancy New File Dialog
image


lusbenjamin:
This sort of achieves https://github.com/Manuel-777/MTG-Arena-Tool/issues/462, but it does not attempt to retrofit the existing log selection dialog that appears during tool launch.

lusbenjamin:
This now also completes https://github.com/Manuel-777/MTG-Arena-Tool/issues/462
image

window_main/settings.js
```diff
@@ -65,6 +66,7 @@ function openSettingsTab(openSection = lastSettingsSection, scrollTop = 0) {
 
   wrap_l.appendChild(createDiv(["list_fill"]));
   wrap_l.appendChild(createDiv(["settings_nav", "sn1"], "Behaviour"));
+  wrap_l.appendChild(createDiv(["settings_nav", "sn7"], "Arena Data"));
```
Manuel-777: I dont think it would hurt to jam "Arena Data" as `sn2` and then pop the rest?

window_background/background.js
```diff
@@ -922,6 +922,7 @@ function onLogEntryFound(entry) {
           default:
             break;
         }
+        pd.settings.log_datetime = entry.timestamp;
```
Manuel-777: Probably not the best use of `pd` ? This data needs to be shared with renderer but not persisted. Same as in `ackground-util.js:50`

lusbenjamin: maybe not, but it does the job quickly. I won't tell if you won't. I moved them into the area of `player-data` that is not persisted.

window_main/settings.js
lusbenjamin: Well I was trying to leave it all alone, but since it bothers you as much as it bothers me...

PR #498, opened 2019-07-22
Move Draft Replays

lusbenjamin:
### Motivation
https://discordapp.com/channels/463844727654187020/518522668212027402/591068303120793622
- Draft Replays are the only items on the "History" page that are not Matches
  - Causes bizarre behavior with some of the filters and lots of weird edge-case logic in the aggregator
  - Confuses users trying to find draft replays

### Approach
- Move Draft Replays to Events page inside the corresponding event
  - always sort draft replay item row to the top
- Clean up History page and aggregator logic
- Rename "History" to "Matches"
- Bonus enhancement: tooltips on main nav icons
- Bonus enhancement: matches on the Events page now show on play/draw
- Bonus bugfix: bugfix draft replays handle drafts without tileID (was https://github.com/Manuel-777/MTG-Arena-Tool/pull/496)

### Demo
image
image

PR #499, opened 2019-07-22
Misc date fns refactors

lusbenjamin:
### Motivation
This PR refactors several places where we manually implemented date-related logic with equivalent (or superior) alternatives from [our `date-fns` library](https://date-fns.org/v2.0.0-alpha.27/docs/Getting-Started). This includes all of #495 without the parts that conflicted with #497 and #498.


Manuel-777:
Is this g2g? I tested and it works but I wonder if you think its stable enough or if there are more stuff to be replaced/refactored

lusbenjamin:
I have been running this locally for almost a week, so all of the affected areas have been pretty well tested. No more major refactors planned, so I'd say g2g!

PR #501, opened 2019-07-25
bugfix force_open_settings index nums

lusbenjamin:
### Motivation
A long time ago, in a PR far, far away...

image

...which has nothing to with why the links to the overlay and privacy settings page are broken in `master`. Nope. Nada.


Manuel-777:
Nada por aqui..

PR #521, opened 2019-08-09
Filters v3 WIP

lusbenjamin:
decks-v3
history-v3


lusbenjamin:
### Explore page moves into section of Decks page
image

lusbenjamin:
### Separate Explore page with side filters
(note that these are simply styled to look like the common `filters` but do not share code yet.)
image

PR #523, opened 2019-08-09
Fixed several bugs and date parsing improvements

AnnanFay:
Fixed bug where unsetting date format was impossible.
Fixed issue of date parsing not working.
Fixed drafts not using timestamps.
Renamed pd to playerData.

More details:

1. Date parsing now works.
2. Setting a config date format value will disable auto-detection
  2b. Added fix to update peoples broken configs where the format is set to "undefined"
3. Date parsing errors will always appear in the log
4. Drafts now use the date parser
5. Match dates now use the date parser
6. Renamed pd (forgot to put this in a different commit)

Reviewers should focus on 2b. in `background.js`, date parsing changes in `labels.js`, `background.js` and `background-util.js`.


window_background/background-util.js
```diff
@@ -116,6 +157,7 @@ module.exports = {
   ipc_send,
   normaliseFields,
   parseWotcTime,
+  parseWotcTimeFallback,
```
lusbenjamin: all external references to `parseWotcTime` were replaced with `parseWotcTimeFallback`. If we don't actually need to expose both fns, we could make the refactor less disruptive by isolating the changes to this module (rename fallback to old name, make strict fn private).

window_background/labels.js
```diff
@@ -816,6 +821,14 @@ function onLabelInDraftDraftStatus(entry, json) {
   };
   data.draftId = data.id;
   data.set = getDraftSet(eventName) || data.set;
+
+  if (!data.date) {
+    // the first time we check the 
+    // draft status we set the date.
+    data.timestamp = entry.timestamp;
+    data.date = logTime;
+  }
```
lusbenjamin: I like that you save the original `timestamp` value as well. (pretty sure won't do anything with it downstream, but I like it anyway)

window_main/history.js
```diff
@@ -266,6 +266,12 @@ function handleOpenDraft(id) {
   });
 }
 
+function localDateFormat(date) {
+  return `<relative-time datetime="${date.toISOString()}">
+    ${date.toString()}
+  </relative-time>`;
+}
+
```
lusbenjamin: I do not have strong opinions about this approach vs `date-fns`, but I am curious to know what motivated this change.

For reference, we also use the custom `local-time` tag on the Economy page, but neither custom tag has yet graduated to a common fn in `util` or `renderer-util`.

window_background/background.js
```diff
@@ -246,12 +247,29 @@ ipc.on("save_app_settings", function(event, arg) {
...
+
+  // Some people's date formats are set to "undefined"
+  // These should be an empty string.
+  if(appSettings.log_locale_format === "undefined") {
+    appSettings.log_locale_format = "";
+    rstore.set("settings", appSettings);
+  }
+  // include more fixes below. Be as specific 
+  // and conservitive as possible.
+}
```
lusbenjamin: Ty ty for creating a minimal process for this!
If at all possible, I would try to add comments describing any known time-bounds for each fix.
e.g. `// bug first introduced in 2.8.4 (2019-07-25)`

window_background/background.js
```diff
@@ -513,14 +531,14 @@ function loadPlayerConfig(playerId, serverData = undefined) {
       }
     })
   };
-  const playerData = {
-    ...pd,
+  const __playerData = {
```
lusbenjamin: 😸 lol name so python

window_background/background.js
```diff
@@ -1222,66 +1240,72 @@ function addCustomDeck(customDeck) {
...
   if (debugLog || !firstPass) store.set("decks." + id, deckData);
 }
 
 //
-function createMatch(arg) {
+function createMatch(json, matchBeginTime) {
   actionLog(-99, new Date(), "");
 
-  currentMatch = _.cloneDeep(currentMatchDefault);
+  var match = _.cloneDeep(currentMatchDefault);
```
lusbenjamin: this is a potentially risky name for a global variable, particularly in the `background` area of the codebase where globals still lurk beneath the surface. Please consider changing to `const` or `let`.

window_background/background.js
```diff
@@ -1414,7 +1438,7 @@ function getColorArchetype(c) {
 }
 
 //
-function getOppDeck() {
+function getOppdeck() {
```
lusbenjamin: Utterly trivial question: why not use camel case here? I kind of expected you to rename this to `getOpponentDeck` if you were going to rename it at all.

window_background/background-util.js
AnnanFay: The name change is important because it now explicitly notes that the parsing code being called has a fallback. I was going to rewrite all code to have error handling, but figured it would be better to just change it to use the `Fallback` function.
New code should preferentially use the non-fallback function and handle their own errors.

window_main/history.js
AnnanFay: I wanted to be able to know exactly when something happened.
image
Before the date fixes if you reloaded an old log it would set them all to `new Date()`, then a few hours later it would be impossible to know when the matches *actually* happened since they all said "n hours ago".
Also just nice for usability, we should probably move all dates to use the code - and move it into `./shared`.

window_background/background.js
AnnanFay: I don't understand what you mean. `match` is not global.  The previous variable `currentMatch` was global - and messing around with globals like it was doing is dangerous.
The new way creates a local variable called `match`, modifies it, then sets the global variable. I didn't feel like mucking around with what is using `currentMatch`.

window_background/background.js
AnnanFay: Damn, that's a mistake. I fucked up the first time I tried replacing `pd` with `playerData`.
`getOppDeck` -> `getOpplayerDataeck` -> `getOppdeck` 😳

window_background/background-util.js
lusbenjamin: 👍

window_main/history.js
lusbenjamin: Wanting the precise tooltip makes sense, also something I missed having before.

window_background/background.js
lusbenjamin: 😅  This was just me misreading your code initially, I didn't see that you were still using `currentMatch` below.

window_background/background.js
lusbenjamin: 🤣

window_background/labels.js
```diff
@@ -790,19 +792,22 @@ function getDraftSet(eventName) {
...
 
 function getDraftData(id) {
-  const draftData = pd.draft(id) || {
+  const draftData = playerData.draft(id) || {
     ...createDraft(),
     id,
     draftId: id,
-    date: new Date(),
-    owner: pd.name
+    date: undefined, // previously new Date()
```
Manuel-777: Wouldn't this cause problems if the user has not set a custom timestamp format and the parser is unable to parse it? what is the result of that happening in case the user has played a couple draft / matches?

AnnanFay: This is not the correct place to set a default date regardless. The correct place to set a default date is the location where date parsing fails.

In `onLabelInDraftDraftStatus` we set the date. If parsing fails because of the format then it will default to the current time (because it's using `parseWotcTimeFallback`).

This also applies to other things. The default values for data should try to avoid masking bugs so using reasonable values is a bad idea. If a fallback value is required for when data is missing then it should be done as late as possible.

Now that I think about it, it might even be a good idea to completely remove the fallback `new Date()` the date parsing and instead add a new field to all data called `parseDate`. This field can then always be set to `new Date()` regardless of what the date of the event is.Then when displaying the date of an event in the UI you check `event.date` and if it's missing you use `event.parseDate`. In this way you completely avoid the issue of accidentally adding invalid dates. `event.date` will *always* be a parsed log timestamp and `parseDate` will always be when the data was parsed.

AnnanFay: You are right about the events though, if we don't get an `onLabelInDraftDraftStatus` event then it could cause problems. Hmm, I suppose we can just pass the date into `getDraftData` and move the `if (!data.date) {` code into that function.

window_background/background.js
```diff
@@ -281,7 +281,12 @@ ipc.on("overlayBounds", (event, index, bounds) => {
 ipc.on("save_user_settings", function(event, settings) {
   // console.log("save_user_settings");
   ipc_send("show_loading");
-  const updated = { ...playerData.settings, ...settings };
+  let refresh = true;
+  if (settings.skip_refresh) {
+    delete settings.skip_refresh;
+    refresh = false;
+  }
+  const updated = { ...pd.settings, ...settings };
```
AnnanFay: Needs to be playerData.

PR #525, opened 2019-08-10
The user-data file should have an explicit name

AnnanFay:
Calling it the UDF / User Data File means we can ask people to upload it, delete it rename it without needing to use euphemisms like "a .json file named with your User ID".

Once people know that the "weird file with a random hexadecimal name" is the UDF then it makes explanations easier.


Manuel-777:
I think we should display the data file name and allow to be copied for users to identify it when something goes wrong, and navigate to it. That could also help with the new naming to "stick" :)

PR #527, opened 2019-08-12
Added some liitle features I personally like

DanielVolz:
Main changes:

- Show/hide hover card in overlay
- Show/hide draw cards odds in overlay
- Resize hover card in overlay and in collection

I'm not a experienced programmer, so there might be some bad code in there.

PR #530, opened 2019-08-14
remove old overlay code

lusbenjamin:
This PR removes the deprecated pre-v3 overlay code.
  - Double checked that there were no code references to "window_overlay"
  - Quickly manually verified things still seem to work as before


Manuel-777:
I was planning to wait a little bit and see if everything works for everyone before removing the old stuff, just in case we need to maintain the old code in parallel (like a "use legacy overlays" option)

But it seems like too much trouble to maintain if we update anything, what do you think?

lusbenjamin:
I was also content leaving it around initially, but was inspired to remove it to prevent confusion for contributors new to the codebase (e.g. #527). If we need to get a copy of the old file we can always use the git history, e.g. `git checkout v2.8.7`

PR #531, opened 2019-08-14
Deck Winrate Confidence Intervals

lusbenjamin:
### Motivation
This PR aims to add a simple secondary metric to help users gauge how "reliable" the primary winrate metric is for each deck. The math is a from-scratch implementation of the Wald Interval. (Warning: this type of math is not my specialty!)

https://www.channelfireball.com/articles/magic-math-how-many-games-do-you-need-for-statistical-significance-in-playtesting/
https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Normal_approximation_interval

### Demo
image
image

### DIFF
--- a/shared/stats-fns.js
+++ b/shared/stats-fns.js
@@ -131,3 +131,16 @@ function hypergeometricSignificance(
   let retVal = math.subtract(1, math.multiply(weightedAverage, 2));
   return returnBig ? retVal : math.number(retVal);
 }
+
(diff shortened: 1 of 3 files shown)


lusbenjamin:
Can manually validate output using 3rd party tools, e.g.
http://epitools.ausvet.com.au/content.php?page=CIProportion&SampleSize=130&Positive=79&Conf=0.95&method=1&Digits=4

PR #532, opened 2019-08-15
Added recording of mythic leaderboard and display in UI.

AnnanFay:
image

### DIFF
--- a/window_background/labels.js
+++ b/window_background/labels.js
@@ -308,8 +308,23 @@ function onLabelInEventGetCombinedRankInfo(entry, json) {
   rank.limited.lost = json.limitedMatchesLost;
   rank.limited.drawn = json.limitedMatchesDrawn;
 
+  rank.constructed.percentile = json.constructedPercentile;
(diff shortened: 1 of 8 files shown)


window_background/labels.js
```diff
@@ -308,8 +308,23 @@ function onLabelInEventGetCombinedRankInfo(entry, json) {
...
+  rank.constructed.seasonOrdinal = json.constructedSeasonOrdinal;
+
+  rank.limited.percentile = json.limitedPercentile;
+  rank.limited.leaderboardPlace = json.limitedLeaderboardPlace;
+  rank.limited.seasonOrdinal = json.limitedSeasonOrdinal;
+
+  var infoLength = Object.keys(json).length - 1;
+  if (infoLength != Object.keys(rank).length) {
+    console.warn("rankInfo is not processing all data.", Object.keys(json));
+  }
```
lusbenjamin: 👍

AnnanFay: It seems like it might be a good idea to have a generic way of saying "This is what we expect the log output to look like - if it adds new fields please warn us." to all fields.
For now I just added this.

Issue #483, opened 2019-07-17
Opening the app disables (systemwide) "option + key" characters in macOS


garvamel:
I'm on a MacBook Pro (Retina, 13-inch, Late 2013) running macOS version 10.14.5. Tool version 2.8.0.

As the title, whenever I open the app, I can't use "option + key" to write special characters systemwide (it even prevents me from entering the email in this app). Normal operation is restored upon closing.

It's not critical, but I guess it's worthwhile to report it. Thank you for this great tool!

garvamel:
I didn't notice the 2.8.1 was out, and it seems to work fine in this one.

Manuel-777:
Cool! It still puzzles me as what could cause it, what version were you using before ?

garvamel:
Just the 2.8.0.

There's this line in `window_overlay/overlay-process.js:113` which might have to do with it. I'll reopen if the issue combes back after a reboot or something. Thanks again!

`globalShortcut.register("Alt+" + (index + 1), () => {
        overlay.webContents.send("close", -1);
      });`

Manuel-777:
I am reopening this as other users reported the same. Upon further investigation this is the expected behaviour of `globalShortcut.register`, which is the only way to trigger shortcuts when the app does not have focus. we should probably choose a less invading shortcut for these and allow for change in the future.

PR #534, opened 2019-08-17
Keyboard shortcuts

Manuel-777:
People who already have keys binded with our default values had issues using tool, sometimes making the overlays completely useless since they had to manually go to the settings page to enable them. Users also experienced default OS actions being binded to our defaults.

image

Tried to make it as easy as possible to add a new shortcut in the future, they should be addded both in `main.js:setSettings()` (as the action/function bind) and `constants.js:SHORTCUT_NAMES`.


lusbenjamin:
Allowed user to specify shortcuts for devtools, and can confirm it is super easy to add new keyboard shortcuts! Also added global setting to enable/disable all shortcuts at once.
image

Manuel-777:
Fixes #483

PR #536, opened 2019-08-18
More mythic rank fixes

AnnanFay:
Fixed a few bugs from the last PR and added saving current rank steps to matches.

Now this works: 
image

### DIFF
--- a/window_background/background.js
+++ b/window_background/background.js
@@ -90,7 +90,12 @@ const {
   onLabelTrackRewardTierUpdated
 } = require("./labels");
 
-const { createDeck, createDraft, createMatch } = require("./data");
+const {
+  createDeck,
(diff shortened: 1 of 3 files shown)


AnnanFay:
@lusbenjamin Nope. I did have a look at that but the changes required to fix it are quite complex. There are several places which assume matches are games.

For an example in the match details we store the players deck and the *totality* of the opponents played cards. So looking at the match details screen - the left side is the full deck list while the right side is cards played in all 1, 2 or 3 games.

This should all really be separate but there is no easy way of storing game-level details. Something like `match.games[1].playedCards = [...]` and `match.games[3].firstPlayer = true`.

All this needs to be done bearing in mind that old data won't have the new structure.

Issue #538, opened 2019-08-18
Multi language databases / database cache


Manuel-777:
We should make databases download only when they need to be updated.
Currently databases are versioned but we cant *check* the database version on startup, a new API endpoint is needed for this.

@AnnanFay 's input:
> Could make the files `./database/en-US/v0.8.1-database.json`, with `./database/version.info` containing the filename string `v0.8.1-database.json`.
> On startup the version is downloaded and checked against the local version. If there's a new version it downloads the correct language specific database.

We should also improve the database creation script to make it faster and output all languages at once with auto versioning.

Manuel-777:
This is mostly done now! (6bc2c795269da8611fa2777234b213155519ba52) We just need to add a dialog in settings to change the desired language in settings.

spphinslove:
Can't wait to use Chinese.
Any plan to make GUI support Chinese?

Manuel-777:
f106ed27fab52e5a045f99c9bfb84dbad93a24e6 now allos to change between different languages in settings.

PR #547, opened 2019-08-20
cache latest metadata

lusbenjamin:
### Motivation
Sometimes our remote hosting service goes down. Sometimes this also happens when the latest stable or beta release depends on the most recent metadata database to run correctly (the distributed source code `database.json` is stale in some way that causes the tool to break). This PR attempts to better handle those scenarios by keeping a cached version of the latest meatadata database alongside the other JSON stores.

### Demo
image

### DIFF
--- a/shared/database.js
+++ b/shared/database.js
@@ -1,9 +1,12 @@
 const path = require("path");
-const { remote, ipcRenderer: ipc } = require("electron");
+const { app, remote, ipcRenderer: ipc } = require("electron");
 const fs = require("fs");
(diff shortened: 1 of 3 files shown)


lusbenjamin:
(probably should ship with #541)

Manuel-777:
Added new api endpoint to get latest metadata version;
`https://mtgatool.com/database/latest/`
So, the link to version 11 for instance would be;
`https://mtgatool.com/database/11.json`

Still no language specific files as per #538 but it opens up that door !

PR #548, opened 2019-08-21
Two cards size slider into one

DanielVolz:
Just made the two cards size slider into one, like @Manuel-777 suggested.
And added a little extra size to overlay hover cards, when the UI scale is reduced.


Manuel-777:
Actually the way its on master is now proper, since this way when you hover on a collection cad the over would be the same size as the card on the collection page :D

DanielVolz:
I think the way it is now in master is not very good (the 2 sliders were my idea in the first place but now I have a better understanding). Because you could change the hover card size of the collection cards even smaller than the actual collection cards. This is not intended and makes no sense. The hover card in the collection should always be bigger than the actual card. It is its purpose.

So in this PR you have one slider but you just change the size of the overlay hover cards and of the collection cards. And not of the collection hover cards. They are always a little bigger (like it was initially. 70px bigger I think), But the hover card will be not be displayed if the card size of the collection cards is bigger than 340px (I added this in my previous PR).

TL:DR
My PR combines both sliders, so you can change the overlay hover card size and the collection size cards but not the overlay card size in the collection.

I hope this makes sense :)

PR #554, opened 2019-08-25
Metadata maker

Manuel-777:
This set of scripts are aimed to completely replace the server sided script we currently use to generate our metadata.

Currently our sever sided script gives no output of possible failures and debugging may become troublesome when tings fail, like downloading the required data we utilize (ranks, metagame, scryfall).
This script now outputs wehenver we download something and shows when the files are generated.

To run use `npm run metadata`

I am still exploring weter this should be a separate package from MTG Arena Tool or not. It is written to be completely independent of the tool.


Manuel-777:
This script is also the beggining of multi language metadata as suggested in #538

PR #588, opened 2019-09-20
Seasonal rank

Manuel-777:
Begin tracking rank by season


resources/amcharts/charts.js
```diff
@@ -0,0 +1 @@
...(this)||this;return t.className="ChordLink",t.middleLine=t.createChild(Ve),t.middleLine.shouldClone=!1,t.middleLine.strokeOpacity=0,t.applyTheme(),t}return r.c(t,e),t.prototype.validate=function(){if(e.prototype.validate.call(this),!this.isTemplate){var t=this.startAngle,i=this.endAngle,n=this.arc,r=this.radius,a=this.dataItem.fromNode,o=this.dataItem.toNode,s=0,l=0;a&&(s=a.pixelX+a.dx,l=a.pixelY+a.dy);var c=0,p=0;if(o&&(c=o.pixelX+o.dx,p=o.pixelY+o.dy),r>0){var d=r*u.cos(t)+s,y=r*u.sin(t)+l,g=r*u.cos(i)+c,f=r*u.sin(i)+p,m={x:0,y:0},v=h.moveTo({x:d,y:y});v+=h.arcTo(t,n,r),v+=h.quadraticCurveTo({x:g,y:f},m),v+=h.arcTo(i,n,r),v+=h.quadraticCurveTo({x:d,y:y},m),this.link.path=n>0?v:"",this.maskBullets&&(this.bulletsMask.path=v,this.bulletsContainer.mask=this.bulletsMask);var x=t+n/2,b=i+n/2,P=this.middleLine;P.x1=r*u.cos(x)+s,P.y1=r*u.sin(x)+l,P.x2=r*u.cos(b)+c,P.y2=r*u.sin(b)+p,P.cpx=0,P.cpy=0,P.stroke=this.fill,this.positionBullets()}}},Object.defineProperty(t.prototype,"radius",{get:function(){return this.getPropertyValue("radius")},set:function(e){this.setPropertyValue("radius",e,!0)},enumerable:!0,configurable:!0}),Object.defineProperty(t.prototype,"arc",{get:function(){return th
```
Lego6245: Not a huge fan of these dependencies living inside of the resource folder instead of as a node module or similar. Is there a reason why you took this approach? (ditto for the other three .js files inside of resources)

Manuel-777: Límited is working, the chart only shows constructed because that's what I was testing. Ideally there should be a toggle to switch between the two.

The js are there because I couldn't get the npm packages to work, another option was to use the URL instead, but I like it to be available right away. They sort of expect to be used in browsers and not in node.

lusbenjamin: Looks like `amcharts` is primarily intended to be used by projects with transpilers (webpack, core-js, require-js or some equivalent). The modules distributed via `npm` will not be compatible with non-ES6 out of the box.

lusbenjamin:
visiting the new "Timeline" page causes later visits to the Decks or History page to have a messed up stats panel:
image

resources/amcharts/charts.js
Lego6245: re: the new inclusion of babel, can we re-visit the inclusion of armcharts as a static asset here?

Manuel-777:
Closing as this is outdated an buggy as hell. I will revisit later.

PR #597, opened 2019-09-27
Correct ELD release date

SabreCat:
I noticed that Throne of Eldraine was sorting to the bottom of the list in the Collection Stats view. After a bit of digging, I believe this was caused by an error in the database JSON that had its release date in 2016. This changes the ELD release date to `2019-10-04` instead of `2016-10-04`, hopefully resolving the sort order.


Manuel-777:
Thanks for noticing! the correct fix is changing the constant in `/scripts/metadata-constants.js` trough, That file then gets "compiled" into the database.js that is downloaded separately from mtgatool. `database.js` is just a "dummy" we provide for the first use of the app, so it does not force users to be online to launch it.

SabreCat:
@Manuel-777 Aha, I see! I've corrected the PR, though at this point it's probably easier for you to just change it in master, haha.

PR #607, opened 2019-10-03
Transparency & click fix

Manuel-777:
This PR aims to fix the transparency and clicking issues in Linux by going back to [electron-transparency-mouse-fix](https://www.npmjs.com/package/electron-transparency-mouse-fix), this time with some better understanding of the problems we are facing.

Only tested on Linux, which was the only platform where the hovering and clicking failed. To merge this it needs to work on Windows as well.

### DIFF
--- a/main.js
+++ b/main.js
@@ -570,7 +570,7 @@ function createOverlayWindow() {
     }
   });
   overlay.loadURL(`file://${__dirname}/window_overlay_v3/index.html`);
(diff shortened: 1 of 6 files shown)


AnnanFay:
Everything seems to work fine.

One interesting thing is that it's now possible to click on other windows during edit mode. I'm not sure if this is a bug or an improvement honestly.

PR #616, opened 2019-10-07
Fix mana symbol tyling issues

Lego6245:
Currently, the mana symbols on the deck page (and i'd imagine elsewhere) tile themselves vertically due to a lack of `background-repeat` css style. Additionally, when resized to be smaller than the 20px width, the box shadow no longer aligns with the mana symbol.

This PR fixes both of these issues, by applying a `background-repeat: no-repeat` and a CSS filter for drop shadow, and removes the outdated CSS.

### DIFF
--- a/shared/shared.css
+++ b/shared/shared.css
@@ -590,11 +590,11 @@ label {
     z-index: 15;
     pointer-events: none;
     background-size: contain;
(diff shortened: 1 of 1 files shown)


Manuel-777:
Good one on the shadows! thanks!

PR #617, opened 2019-10-07
Add "Sort By Wildcards" function

### LINKED ISSUE #566: Sort Decks by Wildcards Needed
Feature suggestion: a dropdown option to sort the My Decks screen by how far off they are from being built. Rare wildcards needed or estimated boosters needed would be the most useful!

Lego6245:
Add the capability for decks to be sorted by # of wildcards needed. This will first sort by total number needed, then by mythics, then by rares, etc. Open to removing the "total number" as the primary sort if we feel it's more organized that way.

Should resolve #566

### DIFF
--- a/window_main/aggregator.js
+++ b/window_main/aggregator.js
@@ -15,7 +15,11 @@ const {
 } = require("../shared/constants");
 const db = require("../shared/database");
 const pd = require("../shared/player-data");
-const { getReadableEvent, getRecentDeckName } = require("../shared/util");
+const {
+  getReadableEvent,
+  getRecentDeckName,
+  get_deck_missing
+} = require("../shared/util");
 const { normalApproximationInterval } = require("../shared/stats-fns");
 
 // Default filter values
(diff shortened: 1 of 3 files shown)


window_main/aggregator.js
```diff
@@ -460,6 +467,33 @@ class Aggregator {
     );
   }
 
+  _sumNumberWildcardsMissing(missingCards) {
+    return Object.values(missingCards).reduce((accum, num) => {
+      if (num) {
+        return accum + num;
+      }
+    });
+  }
```
lusbenjamin: As a possible alternative to using "total wildcards" as the primary sort metric, you could consider importing `util.getBoosterCountEstimate` and using it instead. The result is more of a "sort by booster cost" that falls back on number of missing wildcards when booster estimated cost is identical:
image

Lego6245: Good thing to consider. In your sort example, you can see a consequence of this, where you need 164 rare WCs but only 932 boosters for the second deck, as opposed to 88 Mythic WCs or 1152 boosters.

However, the rate of WC acquisition is such that you'll spend longer getting mythic WCs then you would getting rare WCs (hence the booster count delta) Perhaps this sort could be renamed to something like "By Incomplete" or something more natural sounding to capture that we're assessing how far you are from completing this deck, using these various values as proxies for the overall quantitative assessment?

Manuel-777: I like both methods, sort by boosters required does give a more accurate rate of how far you are from building the deck trough, since it takes into account the rarity of the wildcards and the one you need the most of, rather than adding all of them.

Lego6245: Considering that often times, one's source of wildcards is primarily from opening boosters (with a much more secondary source being vault progress), I think I will refactor this to be a "By Incomplete" check, first checking for # boosters, then # of mythics/rares/uc/c, as those tend to take the longest. I don't think there's a more "perfect" representation of how long it will take, but we can always change it in the future.

PR #624, opened 2019-10-12
[RFC] [DO NOT MERGE] Move to Webpack

Lego6245:
In this PR, I make the necessary (and hopefully minimal) changes to transition the project to build via web-pack.

This PR is not in a mergable state, as CI will not understand what has happened, packaging scripts will need to be updated, etc. However, this is a proof of concept that this move is possible, and I'll outline what I needed to do in order to make it happen.

To try it out, clone my branch, then run `npm run webpack` then `npm-start`


Lego6245:
If this is to be merged in, a few things need to happen.

1. #621 needs to merge
2. A refactor to use ES6 style imports should happen. I don't believe this is contingent on webpack. Part of the changes here include this, but can be replicated out of band by looking at 1140e94 and undoing the changes specific to the webpack configuration changes.

Lego6245:
The changes I needed to make beyond the above that are webpack specific:

1. Generate a webpack-config file (seen in the root)
2. Re-orginize the code into src and static folders (more reorganization could make things cleaner)
3. Take advantage of some webpack features to share common code and resolve assets.

Those alone should basically just be bulk moves and a few more obvious changes I'm right about the prior 3 being review-able sans the webpack changes themselves.

Lego6245:
Changes that still need to be made.

1) Production builds and CI need to be updated to look at the /dist/ folder.

Lego6245:
Closing, this PR is no longer helpful.

PR #625, opened 2019-10-12
[Cleanup] Organize files under src/

Lego6245:
In preparation for bigger work, I've reorginized files so that they sit under /src/

The first two commits are just moves. The last commit contains the changes so that the new directory structure is able to build and run as expected.


Lego6245:
Build seems to be failing lint due to `messages_pb`. Any pointers? Should we add this to the ignored list, since we aren't supposed to be touching the file anyway outside of generation?

PR #627, opened 2019-10-12
Added an Advanced Filter : Owned quantity

thoms999:
Hello Dear Manuel,

The only thing that missed me in your tool was the possibility to filter cards based on the quantity I own. So I did it ! Maybe you will consider it if it looks good for you. 
(i am new to GitHub so sorry if i do something not really well)

Best regards, thank you very much for your work.

> - It works like the CMC filter (I copy-pasted it), you can filter <,> and/or= the owned quantity of the card. It is more precise than the previous filters ("Unowned", "Complete" and "Incomplete"), These ones have been removed.
> - Unowned cards are displayed if filter is null, = 0 or if "Lower than" is checked.
> - Default value is null and "equals to" (like the CMC filter).

owned quantity

### DIFF
--- a/window_main/collection.js
+++ b/window_main/collection.js
@@ -432,18 +432,6 @@ function openCollectionTab() {
 
   addCheckboxSearch(
     cont,
-    '<div class="icon_search_unowned"></div>Show unowned',
-    "query_unown",
-    false
-  );
-  addCheckboxSearch(
-    cont,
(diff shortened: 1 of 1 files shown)


Manuel-777:
Hi there! Thanks for contributing! Changes look sweet, ill do some testing and see if we can merge this.

PR #631, opened 2019-10-14
bugfix collection stats dom timing

lusbenjamin:
### Motivation
https://discordapp.com/channels/463844727654187020/467737642306371584/633239287462363136
image

### DIFF
--- a/window_main/collection.js
+++ b/window_main/collection.js
@@ -760,6 +760,8 @@ function printStats() {
   mainDiv.appendChild(top);
   mainDiv.appendChild(flex);
 
+  openSetStats(stats.complete, "Complete collection");
+
   //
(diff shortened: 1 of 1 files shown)


Manuel-777:
The idea of putting it afterwards was that, if you change a filter, then it would "remember" the last set chosen.

For example;
- Open Collection Stats
- Open any set (Core Set 2020)
- Change "Show Cards" to another value
- Set page reloads, should open Core Set 2020 again. Now instead it resets to "Complete Collection"

lusbenjamin:
That makes sense. We really need a better generic system for attempting to gently cache user state for all filters, display settings, "browser history", etc. Perhaps we can get away with simply increasing the timeout to 500-1000?

PR #638, opened 2019-10-18
Added basic rare drafting info to draft rows

AnnanFay:
image

image

### DIFF
--- a/window_main/economy.js
+++ b/window_main/economy.js
@@ -20,7 +20,8 @@ const {
   formatNumber,
   formatPercent,
   resetMainContainer,
-  toggleArchived
+  toggleArchived,
+  createInventoryCard
 } = require("./renderer-util");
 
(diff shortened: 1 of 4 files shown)


window_main/economy.js
```diff
@@ -662,28 +663,8 @@ function createChangeRow(change, economyId) {
...
-
-      img.addEventListener("click", () => {
-        if (db.card(grpId).dfc == "SplitHalf") {
-          card = db.card(card.dfcId);
-        }
-        //let newname = card.name.split(' ').join('-');
-        openScryfallCard(card);
-      });
+      var divWithNoName = createInventoryCard(card);
+      flexRight.appendChild(divWithNoName);
```
lusbenjamin: I've been through the closure on a div with no name
It felt good to be out of the rain
In the closure you can remember your name
'Cause there ain't null pointer for to give you no pain…

PR #641, opened 2019-10-20
Files under src/

Manuel-777:
Same as #625 but with updated changes (easier to re-do this than updating the code)

@Lego6245 

Only problems I noticed:
- Files under user directory are created under `/Electron` , so old user files are not detected.
- Needed to clear node_modules and do `npm install` again.


package.json
```diff
@@ -55,7 +55,7 @@
     "clean:modules": "rm -fr node_modules",
     "clean:lock": "rm -f package-lock.json",
     "postclean": "npm install",
-    "start": "electron .",
+    "start": "electron ./src/main.js",
```
Lego6245: electron will use your package.json to calculate it's entry point. You've already updated `main` to point to the new directory, this can remain `electron .`

https://stackoverflow.com/questions/35630934/app-getpathuserdata-seems-to-give-the-wrong-path

PR #643, opened 2019-10-20
Adds react support to the project

Lego6245:
Stacked on #642 

This adds support for react to the project, and converts one component to a react component, showing how we can incrementally adopt React.


src/shared/select.js
```diff
@@ -1,136 +0,0 @@
...
-    }
-  });
-
-  parent.appendChild(selectContainer);
-
-  return selectContainer;
-}
-
-export { createSelect, selectAdd };
-function selectAdd(selectElement, callback) {
```
Lego6245: this function was no longer used based on my searching

Manuel-777: Youre correct! This was the first implementation of `reateSelect()`

src/shared/createSelect.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export function createSelect(
+    parent: Element,
+    options: string[],
+    current: string,
+    callback: (option: string) => void,
+    divClass: string,
+    optionFormatter?: (option: string) => string | JSX.Element,
+): JSX.Element {
+    const selectContainer = createDiv(["select_container", divClass]);
```
Manuel-777: the select_container div is duplicated on line 32, I think render() should point to the parent directly?
This causes all containers to have the same fixed width.

Manuel-777: Ultimately I believe the best approach is using a class instead of a function alone, and adding state to handle things like the container `width`.
Same for any other components we add later on.

Lego6245: I'm not sure I follow the second part. Why do components need to be stateful to handle container width? I think you're right (I'll fix the duplicated <div> problem), but I'm confused as to what you're implying about class based components over functional components (there's very little a class based component can do that a functional component can't, and functional components are the direction the React folks are moving in)

Manuel-777: Im probably outdated with the latest React developments so ignore me if im being dumb :P
Should we then use a prop to give it width?

Lego6245: I'm not sure what this width thing is about, I guess. Is it rendering wrong? I just tried live vs this branch in multiple panes at a variety of window sizes and both are rendering the same.

Lego6245: image

image

Lego6245: I think you could argue that the padding inside each has regressed slightly (not as much whitespace)?

Lego6245: I think I see what's going on. Some parents are manipulating the children to have certain widths.

Right now, the reason why we create the select container outside of the component is 2 fold:

1) To give us a blank space in which to call ReactDOM.render()

2) To give us something that we can return that matches the current expected dom interface consumers are used to.

As more and more pieces of the app become react, we need this function less and less. This function is basically our bridge between the React world of tomorrow and the current reality today. So, hopefully we get to the point where we don't need to do this.

Manuel-777: image
In Explore its wrong particularly, also I see "All Tags" colliding with "All decks" (both having full width)

Lego6245: My most recent pushes fix that problem:

image

Lego6245: I think this is definitely something that we can resolve the more we "reactify" our app, as it seems like we do a lot of hacky manual layout-ing.

Manuel-777: Yes, I noticed that we need that "bridge", so I suggested rendering react directly in the parent div instead of creating an intermediate one. Then pass width or any other css attributes to the JSX Elements in the function. I dont know how that can be done trough, give CSS properties to elements in the React object (thats why I suggested using state to handle it after creating the element outside the constructor). But maybe theres a better way of doing it?

Lego6245: Most of our potential components are wrapped in a div, so the method I'm taking is to abstract the root layer div out as the target for ReactDOM.render, and returning that container. We can't call ReactDOM.render on the parent, as part of the side effects of ReactDOM.render is to make that parent element empty, which causes the entire thing to not render as expected. I think for now to minimally churn anything dependent on createSelect this current pattern works best.

Manuel-777: I see.. Hopefully when more components are added we can do this properly trough! the empty tag feels hacky :)

Lego6245: Yes, any time we reactify something currently, unless we also verify that nobody is manipulating that created component externally, we have to do something like this. At least at every boundry between the React world and the non-React world.

PR #647, opened 2019-10-21
Add Typescript-ESLint

Lego6245:
Stacked on #642 

Adds typescript-eslint, enables some rules, auto fixes some files, manually fixes a few others, leaves some rules disabled because they're too loud right now.


Lego6245:
To aid in review, a lot of these changes were automated with eslint --fix, including most of the const/var stuff. There were a couple of manual fixes done, including some re-ordering of files.

src/window_background/globals.ts
```diff
@@ -6,45 +6,45 @@ import * as mtgaLog from "./mtga-log";
 // Hopefully we'll be able to get rid of all of the ones that can change,
 // and put them into stores or better structures than a giant export list.
 
-let actionLogDir = "";
+const actionLogDir = "";
```
geppy: `globals.actionLogDir` is a mutable binding, otherwise it'd always be `""`.
```suggestion
let actionLogDir = "";
```

src/window_background/globals.ts
```diff
@@ -6,45 +6,45 @@ import * as mtgaLog from "./mtga-log";
...
 
-let logUri = mtgaLog.defaultLogUri();
+const logUri = mtgaLog.defaultLogUri();
```
geppy: This is also a mutable binding... most of these globals are assigned to outside of this file.

src/window_background/http-api.js
```diff
@@ -428,11 +418,6 @@ export function httpBasic() {
                   text: parsedResult.error,
                   time: 10000
                 });
-              } else if (_headers.method == "tou_check") {
```
geppy: Was this just removed to prevent compiler errors?

If so, can the unused local be replaced with a `// TODO` comment?

src/window_background/labels.js
```diff
@@ -25,13 +25,12 @@ import {
 import actionLog from "./actionLog";
 import addCustomDeck from "./addCustomDeck";
 import { createDraft, createMatch, completeMatch } from "./data";
-
-var logLanguage = "English";
+import * as httpApi from "./http-api";
```
geppy: Is there a reason to use a namespace import?  The identifiers are all prefixed with `http`.

```suggestion
import {
  httpSetDraft,
  httpSetMatch,
  httpSetMythicRank,
  httpSetSeasonal,
  httpSubmitCourse,
  httpTournamentCheck
} from "./http-api";
```

src/window_background/updateDeck.js
```diff
@@ -5,10 +5,10 @@ import forceDeckUpdate from "./forceDeckUpdate";
 import getOpponentDeck from "./getOpponentDeck";
 import { objectClone } from "../shared/util";
 
-var lastDeckUpdate = new Date();
+const lastDeckUpdate = new Date();
```
geppy: `update_deck` *should* update this later on, it's just buggy.

Making it `const` makes the bug seem intentional.

```suggestion
let lastDeckUpdate = new Date();
```

src/window_background/updateDeck.js
```diff
@@ -5,10 +5,10 @@ import forceDeckUpdate from "./forceDeckUpdate";
...
 
 function update_deck(force) {
-  var nd = new Date();
+  const nd = new Date();
```
geppy: Inside the branch, I'd expect `lastDeckUpdate = nd;`.

src/window_updater/updater.js
```diff
@@ -1,8 +1,8 @@
 import { ipcRenderer as ipc, remote } from "electron";
+import { openNewGitHubIssue, debugInfo } from "electron-util";
```
geppy: Making these imports unconditional is a behavioral change (which seems unintentional and out of scope for this PR).

Lego6245:
Hey @geppy, thanks for your review. A lot of the changes you're calling out were done automatically via eslint -fix in order to get the build "green" again. It looks like for dual `.js / .ts` projects the `const` rule is perhaps being calculated incorrectly.

I think the proper course of action, at least for this PR, is to separate out those changes into a separate PR, and change this PR to focus soley on the toolchain aspects, while disabling all the rules.

geppy:
That sounds like a solid plan to me.

Lego6245:
Closing in lieu of #664 being the main root

src/window_background/labels.js
Manuel-777: No reason really, its a leftover from when the file was created and used as exported functions.

Issue #563, opened 2019-09-03
Hover card display can get stuck on screen when mousing between displays


SabreCat:
* Have two displays
* Bring up an overlay and pin it against the inner edge of one display
* Mouse over a card name to show its image
* Mouse to the other display
* Hide the overlay

The hover card gets stuck on screen, remaining until you bring the overlay back up, mouse over something else, and mouse away without transitioning between displays in the process.

A simple fix would be to end all card hovers when hiding an overlay. The hover might still get stuck temporarily, but without the overlay-closing issue it would be a minor nuisance to get rid of with a bit of mousework.

PR #652, opened 2019-10-21
Overlay Bugfixes

lusbenjamin:
### Motivation
This is a bundle of misc bugfixes for the overlays:
 - workaround for #563 by adding a 10-second auto-hide to hover card display
 - bugfix: on windows, fixes an issue with the transparency click-through for the first time overlays appear by calling `overlay.setIgnoreMouseEvents(true, { forward: true });`. This affected matches and drafts. Since this is unsupported in Linux, this PR limits the call to non-linux OS.
  - bugfix: on windows, lets the user scroll the decklist/sideboard
  - refactor: standardizes when the `overlay` window is created in `main` process to eliminate the need for null-checks and certain rare race conditions

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -19,3 +19,4 @@ coverage
 # production
 build
 dist
+lib
--- a/src/window_overlay_v3/overlay.js
+++ b/src/window_overlay_v3/overlay.js
@@ -1130,7 +1130,7 @@ ready(function() {
       <div class="outer_wrapper elements_wrapper">
         <div class="overlay_deckname"></div>
         <div class="overlay_deckcolors"></div>
-        <div class="overlay_decklist"></div>
+        <div class="overlay_decklist click-on"></div>
         <div class="overlay_clock_container">
             <div class="clock_prev click-on"></div>
             <div class="clock_turn"></div>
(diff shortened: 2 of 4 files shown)


lusbenjamin:
There are varied reports in the Discord about having the overlays "not allow clicks through". I am uncertain if this PR will have any broader affect on any of those cases.

Manuel-777:
It works well in Linux, only "Issue" I noticed is when using transparent overlays, you cant click trough the empty areas. But I can totally live with that if it fixes the rest of our issues!

.prettierignore
```diff
@@ -19,3 +19,4 @@ coverage
 # production
 build
 dist
+lib
```
Lego6245: thanks for this :D

src/window_overlay_v3/overlay.js
```diff
@@ -1130,7 +1130,7 @@ ready(function() {
       <div class="outer_wrapper elements_wrapper">
         <div class="overlay_deckname"></div>
         <div class="overlay_deckcolors"></div>
-        <div class="overlay_decklist"></div>
+        <div class="overlay_decklist click-on"></div>
```
Lego6245: I assume this css already exists?

Manuel-777: Yes, this class is used by the `electron-transparency-mouse-fix` and handles click-trough and transparency events.

PR #661, opened 2019-10-26
Show winrate interval in explore tab

Thaoden:
Decks in the Explore tab often show a high winrate while having been played only a few matches, which makes it difficult to properly assess their performance.

This PR uses the code that was already present for the My Decks tab to display the winrate interval next to the wins : losses ratio, together with the interval boundaries in the tooltip. The part where it says "more matches are needed" is left out, as it is usually not in the capabilities of the one using the Explore tab to play more matches with the deck (s)he's inspecting.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -725,7 +725,6 @@ export function formatRank(rank) {
   return `${rank.rank} ${rank.tier}`;
 }
 
-
-export function roundWinrate(x){
+export function roundWinrate(x) {
   return Math.round(x * 100) / 100;
 }
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -724,3 +724,8 @@ export function formatRank(rank) {
   }
   return `${rank.rank} ${rank.tier}`;
 }
+
+
+export function roundWinrate(x){
+  return Math.round(x * 100) / 100;
+}
--- a/src/window_main/decks.js
+++ b/src/window_main/decks.js
@@ -213,7 +213,10 @@ export function openDecksTab(_filters = {}, scrollTop = 0) {
       if (dwr.total >= 20) {
         // sample Size is large enough to use Wald Interval
         interval = formatPercent(dwr.interval);
-        tooltip = formatWinrateInterval(formatPercent(dwr.winrateLow), formatPercent(dwr.winrateHigh))
+        tooltip = formatWinrateInterval(
+          formatPercent(dwr.winrateLow),
+          formatPercent(dwr.winrateHigh)
+        );
       } else {
         // sample size is too small (garbage results)
         interval = "???";
--- a/src/window_main/decks.js
+++ b/src/window_main/decks.js
@@ -16,6 +16,7 @@ import StatsPanel from "./stats-panel";
 import { openDeck } from "./deck-details";
 import {
   formatPercent,
+  formatWinrateInterval,
   getTagColor,
   getWinrateClass,
   hideLoadingBars,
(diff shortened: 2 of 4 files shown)


src/window_main/renderer-util.js
```diff
@@ -760,6 +760,11 @@ function formatPercent(value, config = { maximumSignificantDigits: 2 }) {
   });
 }
 
+function formatWinrateInterval(lower, upper){
```
Lego6245: If you can make this `export function` and delete line 1078, that'd be great, as I'm meaning to get rid of the bulk export blocks at the end of files.

src/window_main/decks.js
```diff
@@ -212,10 +213,7 @@ export function openDecksTab(_filters = {}, scrollTop = 0) {
       if (dwr.total >= 20) {
         // sample Size is large enough to use Wald Interval
         interval = formatPercent(dwr.interval);
-        tooltip = `${formatPercent(dwr.winrateLow)} to ${formatPercent(
-          dwr.winrateHigh
-        )} with 95% confidence
-(estimated actual winrate bounds, assuming a normal distribution)`;
```
Lego6245: do we care about losing this text?

src/shared/util.js
```diff
@@ -724,3 +724,8 @@ export function formatRank(rank) {
   }
   return `${rank.rank} ${rank.tier}`;
 }
+
+
+export function roundWinrate(x){
```
Lego6245: I think this makes sense as a util, would like to be spot-checked.

Lego6245:
Your build failed because of prettier, try running `npm run format` and you should get a green build.

src/window_main/explore.js
```diff
@@ -613,13 +618,25 @@ function deckLoad(_deck, index) {
...
   );
+  if (_deck.mt >= 20) {
+    // sample Size is large enough to use Wald Interval
+    d.title = formatWinrateInterval(
+      roundWinrate(deckWinrate.winrate - deckWinrate.interval),
+      roundWinrate(deckWinrate.winrate + deckWinrate.interval)
+    );
+    d.innerHTML += `<i style="opacity:0.6;"> &plusmn; ${formatPercent(
+      roundWinrate(deckWinrate.interval)
+    )}</i>)`;
```
lusbenjamin: may be an extra trailing `)` in the backtick here?

src/window_main/explore.js
```diff
@@ -613,13 +618,25 @@ function deckLoad(_deck, index) {
...
+
+  let colClass = getWinrateClass(deckWinrate.winrate);
   d = createDiv(
     ["list_deck_record"],
-    `${_deck.mw}:${_deck.ml} <span class="${colClass}_bright">(${Math.round(
-      (100 / _deck.mt) * _deck.mw
-    )}%)</span>`
+    `${_deck.mw}:${_deck.ml} <span class="${colClass}_bright">(${formatPercent(
+      deckWinrate.winrate
+    )})</span>`
```
lusbenjamin: my bad, the extra is prolly this one

src/window_main/decks.js
lusbenjamin: I thought the text just moved to the helper fn

Lego6245: oh i see now, my brain must have skipped over it. ignore this, then :)

src/window_main/explore.js
Lego6245: If you're using vscode, colored bracket pairs extension is a godsend :)

src/window_main/renderer-util.js
Thaoden: Will do.

src/window_main/explore.js
Thaoden: Still getting to grips with VS Code, the "whole" IDE VS has it all baked in...

Yeah, the `)` is a leftover from a copied `createDiv()`...

Thaoden:
> Overall, LGTM! Thanks for contributing! I'm glad you updated the "my decks" tab too as part of this PR.

Well, yeah, no use in introducing a utility function and not using it ;)

Just pushed a new version.

src/window_main/explore.js
Thaoden: Alright, I'm completely lost now...

There was an extra `)` on line 638, that one is fixed.

The first bracket on line 628 closes the opening one from `formatPercent(`. The second one is plaintext (not interpreted) and closes the opening plaintext bracket just after `bright">`, all in all, it gives `( ${formatPercent(foo)} )`, so the bracket here is needed.

Manuel-777:
Looks good!

Thaoden:
Happy to see this merged :)

PR #662, opened 2019-10-26
Batch load fixes

AnnanFay:
A fix for working around log files which stall, if there are no messages in 30 seconds then it terminates and moves on.

Now skips past logs which are non-detailed. 

And should now be working with the new babel compilation.

### DIFF
--- a/scripts/load-logs.js
+++ b/scripts/load-logs.js
@@ -4,7 +4,8 @@
 // Usage: Edit `batch_load_files.json` to contain the absolute
 // paths to all backed up log files
 // Run `npm run load-logs` and watch the text scroll by for a few hours.
-// Backup your config file before you do this.
+
+// IMPORTANT: Backup your config file before you do this.
 
 // The code below works by instantiating the background process and
 // sending the minimal necesary IPC calls so it thinks it's starting up as normal.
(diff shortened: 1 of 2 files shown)


scripts/load-logs.js
```diff
@@ -30,11 +35,12 @@ function createBackgroundWindow() {
       nodeIntegration: true
     }
   });
-  win.loadURL(`file://${__dirname}/../window_background/index.html`);
+  win.loadURL(`file://${__dirname}/../lib/window_background/index.html`);
```
Manuel-777: this sould be under `/src`;
```suggestion
  win.loadURL(`file://${__dirname}/../src/window_background/index.html`);
```

Lego6245: arguably you want the transpiled version? though it's not obvious that it's a dependency...

Manuel-777: So pre-running the transpiler is a requirement to run scripts, should we force it before running the script?

AnnanFay: I tried `/src/` at first and it didn't work. If that change is made without anything else then it will break. I am not familiar enough with babel to know the correct other changes to make.

Lego6245:
There is definitely a case to be made that you shouldn't need to open the index.html of the window in order to do any of this, and that scripts should be able to execute without relying on any of the main source code (and if it does, we abstract those pieces out)

PR #664, opened 2019-10-26
Enable Typescript + Eslint

Lego6245:
Split from #647 to contain just the minimal changes require to turn it on. Any rules that caused errors when enabled are set to warn currently. We should try progressively to turn on each rule (which was attempted in #647 but caused the PR to have unintended changes.


.eslintrc.json
```diff
@@ -1,6 +1,5 @@
 {
   "extends": ["eslint:recommended", "plugin:prettier/recommended"],
-  "plugins": ["prettier"],
```
Lego6245: This... actually conflicts with the plugin:prettier/recommend extension, and causes eslint to do nothing. Removing this actually enables prettier + eslint. We should further discuss if it makes sense to also run stand alone prettier or if we should hand the process over to eslint to manage prettier + other lint rules.

.eslintrc.json
```diff
@@ -20,6 +19,7 @@
     "no-global-assign": "warn",
     "no-empty": "warn",
     "complexity": ["warn", 40],
-    "max-statements": ["warn", 100]
+    "max-statements": ["warn", 100],
+    "prettier/prettier": "warn"
```
Lego6245: This disables the rules that are firing on eslint but aren't fixed automatically by prettier itself.

src/shared/card-hover.js
```diff
@@ -145,6 +145,8 @@ export function attachDraftRatings(card, ratingsContainer) {
   let rankValues = card.rank_values;
   let rankControversy = card.rank_controversy;
 
+  // TODO: Refactor and enable this rule.
+  // eslint-disable-next-line prefer-spread
   let maxValue = Math.max.apply(Math, card.rank_values);
```
Lego6245: This is the only occurrence of this rule violation, and should be done first.

.eslintrc.json
geppy: Per [prettier's <cite>Integrating with Linters</cite>](https://prettier.io/docs/en/integrating-with-linters.html), extending `plugin:prettier/recommended` q cite="https://prettier.io/docs/en/integrating-with-linters.html#recommended-configuration">configures both `eslint-plugin-prettier` and `eslint-config-prettier` in a single step</q>.

It shouldn't break itself when it's enabled twice, but not putting it in plugins is their [<q>recommended configuration</q>](https://prettier.io/docs/en/integrating-with-linters.html#recommended-configuration).

package.json
```diff
@@ -59,7 +59,8 @@
     "postinstall": "install-app-deps",
     "format": "prettier --write \"src/**/*.js\"",
     "test": "npm run --silent test:eslint && npm run --silent test:prettier && npm run --silent test:jest",
-    "test:eslint": "eslint . --quiet",
```
geppy: Are we using eslint with other extensions, or can we move this to `.eslintrc.json`?

src/.eslintrc.json
```diff
@@ -0,0 +1,17 @@
+{
+  "parser": "@typescript-eslint/parser",
+  "extends": ["plugin:@typescript-eslint/recommended", "prettier/@typescript-eslint"],
```
geppy: Why is this separate from `../.eslintrc.json`?  Just so `/scripts` and config files are handled separately?

src/shared/card-hover.js
geppy: The rule is asking for this edit:
```suggestion
  let maxValue = Math.max(...rank_values);
```
It runs more code, but the rule is trying to establish spread as the preferred idiom.

package.json
```diff
@@ -59,7 +59,8 @@
...
+    "test:eslint": "eslint . --quiet --ext .js,.ts,.jsx,.tsx",
+    "eslint:fix": "eslint . --quiet --ext .js,.ts,.jsx,.tsx --fix",
```
geppy: Does this fix warnings, or only errors?

.eslintrc.json
Lego6245: Yeah, what I meant was previously none of the prettier violations were being reported through eslint, only through the prettier CLI

src/.eslintrc.json
Lego6245: Yes, since /scripts are currently written in pure .js right now.

package.json
Lego6245: --fix should ONLY hit errors, AFAIK.

package.json
Lego6245: Let me make sure that there isn't an unintended side effect from doing so, and then I will.

Lego6245: Is there something that says `--quiet` can be a config parameter? Only documentation I can find refers to its CLI usage, so we can't move this to a config.

PR #665, opened 2019-10-27
Make the filter panel react-y

Lego6245:
This change makes the filter panel react aware. This is a big of a big one, broken up into mulitple parts:

1) Abstract the re-usable pieces out
2) Move the inline applied css styling into actual css
3) Reorganize the render method to remove duplicated code.

I've done some testing, but extra testing appreciated :)


Manuel-777:
So far it looks good, only thing I noticed is the "archived" button does not work.

Lego6245:
Merged a bunch of changes that should fix that archived button issue. I relish the day the whole pane is React

PR #667, opened 2019-10-27
Consolidate all the share dialog cases

Lego6245:
We have 3 usages of the share dialog right now, and all 3 look about the same. Literally, the only thing that's different is the callback when clicked. So, abstract it out and we're good to go.

Eventually, once the consumers are more "react aware" we can change it so that this is a component that renders a button, but this is the most expedient right now.


Lego6245:
Problem was I was using createInput instead of createDiv. Quick fix :D

PR #669, opened 2019-10-27
Make the Economy page React aware

Lego6245:
Reorginizes the whole economy page, makes it react aware, and hopefully breaks nothing.

Still need to test the archive function, but everything else should be organized properly.


src/window_main/EconomyRow.tsx
```diff
@@ -0,0 +1,347 @@
...
+      return aggregator;
+    }
+  , []) : undefined;
+
+  const checkSkins = checkSkinsAdded && change.delta.artSkinsAdded !== undefined;
+  const skinsToCards = checkSkins ? change.delta.artSkinsAdded.map((obj: { artId: string }) => db.cardFromArt(obj.artId)) : undefined;
+  return (
+    <div className={"tiny_scroll list_economy_awarded"} id={economyId}>
+      {fullContext === "Pay Event Entry" && <EconomyIcon title={"Event Entry"} className={"economy_ticket_med"} />}
+      {checkGemsEarnt && change.delta.gemsDelta && <EconomyValueRecord iconClassName={"economy_gems_med"} title={"Gems"} deltaContent={formatNumber(Math.abs(change.delta.gemsDelta))} />}
```
Lego6245: part of me thinks there's a cleaner way to do this.... need to puzzle on it a bit

src/window_main/EconomyValueRecord.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+function DeltaLabel(props: { content: string, smallLabel?: boolean }) {
+  return (
+    <div className={"economy_sub" + (props.smallLabel ? " small" : "")} >
+      {props.content}
+    </div>
+  );
+}
+
+function UpConta() {
```
Lego6245: may get rid of these two, they don't do much

src/window_main/DateFilter.tsx
```diff
@@ -6,24 +6,25 @@ export interface DateFilterProps extends ReactSelectProps {
...
 export default function DateFilter(props: DateFilterProps) {
     const [ showArchived, setShowArchived ] = React.useState(props.showArchivedValue);
+    const { onArchiveClick, prefixId, showArchivedFilter, ...selectProps } = props;
     
     const onClickArchiveCheckbox = React.useCallback((event: React.ChangeEvent<HTMLInputElement>) => {
       const newValue = event.currentTarget.checked;
       if (newValue != showArchived) {
         setShowArchived(newValue);
-        props.onArchiveClick && props.onArchiveClick(newValue);
+        onArchiveClick && onArchiveClick(newValue);
```
lusbenjamin: Is this a potential location for `!!`? Trying to understand when you choose to use that syntax.

src/window_main/DateFilter.tsx
```diff
@@ -6,24 +6,25 @@ export interface DateFilterProps extends ReactSelectProps {
...
       }
     }, [props.onArchiveClick, showArchived]);
     
-    const { prefixId, options, current, callback, showArchivedFilter } = props;
     return (
-      <div className={"dateCont"}>
+      <div className={props.className + " dateCont"}>
         <div className={"select_container filter_panel_select_margin " + prefixId + "_query_date"}>
-            <ReactSelect options={options} current={current} callback={callback} />
+          <ReactSelect {...selectProps} />
```
lusbenjamin: I like the double use of the destructuring syntax with `...selectProps` to pass through props! Definitely going to try this out the next chance I get.

src/window_main/EconomyDayHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+import EconomyValueRecord from "./EconomyValueRecord";
+import LocalTime from "../shared/time-components/LocalTime";
+
+function localDayDateFormat(date: Date) {
+  return (
+    <LocalTime datetime={date.toISOString()}
+      year={"numeric"}
+      month={"long"}
+      day={"numeric"}>
+    </LocalTime>
```
lusbenjamin: just curious, is the `year={"numeric"}` syntax required because of the wrapper hack, or could you have gotten away with `year="numeric"`?

src/window_main/economyUtils.ts
```diff
@@ -0,0 +1,196 @@
...
+  "Vault.Complete": "Vault Opening",
+  "PlayerReward.OnMatchCompletedWeekly": "Match Rewards: Weekly Wins",
+  "PlayerProgression.OrbSpend": "Orb Spend",
+  "Track Progress": "Match Rewards: Level Up",
+  "Track.RewardTier.Updated": "Mastery Pass Purchase",
+  "Player Rewards": "Match Rewards",
+  "Event Prize": "Event Rewards"
+};
+
+type EconomyTransactionKeys = keyof typeof economyTransactionContextsMap;
```
lusbenjamin: 😖 <-- mfw I first encounter this syntax

src/window_main/EconomyRow.tsx
```diff
@@ -0,0 +1,377 @@
...
+}
+
+function FlexBottom(props: FlexBottomProps) {
+  const { fullContext, change, thingsToCheck } = props;
+  const { checkGemsPaid, checkGoldPaid } = thingsToCheck;
+  return (
+    <div className={"flex_bottom"}>
+      {fullContext === "Booster Open" ? change.delta.boosterDelta.map((booster: any) => <BoosterDelta booster={booster} key={booster.collationId} />) : fullContext === "Redeem Wildcard" ? (
+        <>
+          {!!change.delta.wcCommonDelta && <WildcardEconomyValueRecord count={change.delta.wcCommonDelta} title={"Common Wildcard"} className={"wc_common"} smallLabel={true} />}
```
lusbenjamin: curious about this `!!` syntax convention and when it is better to use. some kind of defensive pattern to ensure `boolean` typing?

src/window_main/EconomyRow.tsx
```diff
@@ -0,0 +1,377 @@
...
+      }
+      return aggregator;
+    }
+  , []) : [];
+
+  const checkSkins = checkSkinsAdded && change.delta.artSkinsAdded !== undefined;
+  const skinsToCards = checkSkins ? change.delta.artSkinsAdded.map((obj: { artId: string }) => db.cardFromArt(obj.artId)) : undefined;
+
+  const xpGainedNumber = change.xpGained && parseInt(change.xpGained);
+  console.log(change);
```
lusbenjamin: nit: `console.log`

src/window_main/EconomyDayHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+    gridArea: "1 / 1 / auto / 2",
+    lineHeight: "64px"
+  };
+
+  return (
+    <>
+      <div style={gridTitleStyle} className={"flex_item gridTitle"}>
+        {getDayString(daysago, timestamp)}
+      </div>
+      <EconomyValueRecord containerDiv iconClassName={"economy_card"} className={"gridCards"} deltaUpContent={formatNumber(dayList[daysago].cardsEarned)} title={"Cards"}/>
```
lusbenjamin: what is `containerDiv`?

src/window_main/EconomyRow.tsx
```diff
@@ -0,0 +1,377 @@
...
+import { cardHasType } from "../shared/card-types";
+import ReactDOM from "react-dom";
+import LocalTime from "../shared/time-components/LocalTime";
+import { string } from "prop-types";
+
+function localDateFormat(date: Date) {
+  return (
+    <LocalTime datetime={date.toISOString()} month={"short"} day={"numeric"} hour={"numeric"} minute={"numeric"} />
+  );
+}
```
lusbenjamin: *cues the Inception noise*

src/window_main/DateFilter.tsx
Lego6245: It's a very clean pattern that makes things much more readable since you're only interacting directly with the props you care about.

src/window_main/EconomyDayHeader.tsx
Lego6245: You don't technically need curly braces for props that are just strings, but I use them everywhere because it's a consistent pattern

src/window_main/economyUtils.ts
Lego6245: Yeah it's a little weird, but hopefully you figured it out.

`keyof`: Give me a string union of the keys.
`typeof`: Give me the type that represents this untyped object

So, you end up with a type representing the string union of the keys presence in the type of `economyTransactionObject`. This way, if we every modify the above const, we are updating the list of valid types elsewhere.

src/window_main/EconomyRow.tsx
Lego6245: So, as you're probably aware, certain values are falsy in javascript. Those certain values include `0`. The problem is if you do some statement like `0 %% "foo"` in Javascript, the resulting value is not `false` but rather `0`. `0` is a valid value to render though, so you end up rendering a `0` instead of nothing, which is intended.

`!!` as a syntax does exactly what you think it does. It essentially asserts that this value is truthy, returning `true` if it is and `false` if it isn't. We coerce this value into being explicitly `false` so if it's `0` nothing will render.

src/window_main/DateFilter.tsx
Lego6245: We don't necessarily care what the the value of this expression is (since we aren't assigning it to a variable), so `!!` would be excessive here.

src/window_main/EconomyDayHeader.tsx
Lego6245: `containerDiv` is a prop. Since it doesn't have an explicit value, it is by default evaluated to `true` in `props.containerDiv`.

lusbenjamin:

PR #670, opened 2019-10-28
size overlay based on the display it's on

geppy:
Overlays are always sized to fit the primary display, which doesn't make sense if the overlay's on a different display.

This fixes that, and delays showing overlays until *after* they're resized.


Manuel-777:
Great one!

PR #678, opened 2019-10-29
Fix eplore tab

Thaoden:
This is a fix for the explore tab.

Previously, `getInputValue` would attempt to read the `value` of the `explore_query_event` _`div`_, which would return undefined and would result in a db error on the server side. 

What we need is the `textValue` of the `button` inside that `div`.


Thaoden:
Actually, this is not quite done yet...

PR #684, opened 2019-11-01
Shared code TypeScript migrations

Thaoden:
A first attempt at TypeScript-izing some of the background stuff, introducing some types, typing the return values of some functions as well as typing their input-values. Did I mention Types?

I _think_ all functions on `database.ts` are working as expected, as far as I can see, neither compile-time nor runtime errors were thrown.

I'm not sure how far spread all the different types should be across files. Currently, `Metadata.ts` includes also the `RewardsDate`, as I didn't want to create a new file with only 4 lines in it. But that's open for discussion.

I'm planning to add more files to this little by little. Feedback is welcome!


src/shared/database.ts
```diff
@@ -0,0 +1,264 @@
+import path from "path";
```
Lego6245: It bus me that this file wasn't picked up as a move.

src/shared/types/Season.ts
```diff
@@ -0,0 +1,42 @@
...
+    Bronze,
+    Silver,
+    Gold,
+    Platinum,
+    Diamond,
+    Mythic
+  }
+  
+  interface SeasonRewardInfo {
+    image1: string | null,
```
Lego6245: nit: i'm not sure semantically null is correct here. we probably want undefined. typescript has the `?` operator shorthand for this. so, `image1?: string` will type as `string | undefined`

src/shared/database.ts
Thaoden: Not sure I understand?

src/shared/constants.ts
```diff
@@ -1,3 +1,16 @@
+interface EnumType {
+  CounterType: Record<number, string>,
```
Lego6245: Record types are great! They tend to be used for situations where you have a string enum or some similar limited subset of type `K` in `Record<K, T>` that you want to restrict, but it's usage here is fine.

src/shared/types/Season.ts
Thaoden: The game output has either string or `null` as value - I didn't want to change the type, but to stay as close to the original as possible. refactorings can be made more safely after the move to TS.

Thaoden:
Now if only I could get this test to run...

src/shared/constants.ts
Thaoden: Better use a dictionary then?

I thought about migrating some of the `const`s that all begin with the same name (`OVERLAY_XXX`, `SETTINGS_XXX`, `MAIN_XXX`) to an `enum` at a later step.

src/shared/constants.ts
```diff
@@ -630,17 +643,17 @@ export const FORMATS = {
...
+export const RANKS: string[] = [
   "Bronze",
   "Silver",
   "Gold",
   "Platinum",
   "Diamond",
   "Mythic"
 ];
-export const CONSTRUCTED_EVENTS = ["Ladder", "Traditional_Ladder"];
+export const CONSTRUCTED_EVENTS: string[] = ["Ladder", "Traditional_Ladder"];
```
Lego6245: nit: we can actually type this pretty well with typescript 3.4 (this is like, hot and fresh typescript)
```
export const CONSTRUCTED_EVENTS = ["Ladder", "Traditional_Ladder"] as const;
```
What this does is informs the typescript compiler to assume the strictest possible typing for this array. This means that elsewhere, if we, say, `map` or `switch` over code this array, or arrays we can define as being members of this type, we don't need default cases if we're exhaustive. I think we should leverage this if we can.

src/shared/constants.ts
```diff
@@ -630,17 +643,17 @@ export const FORMATS = {
...
 export const OVERLAY_FULL = 0;
```
Lego6245: nit: maybe just make this into an enum?

src/shared/constants.ts
Lego6245: See this article from the [3.4 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions)

Lego6245: This goes for pretty much any of these arrays we consider immutable in this file.

src/shared/constants.ts
Lego6245: Nope, you're fine here. No action needed.

src/shared/database.ts
Lego6245: You see how constants.js got marked as a move and edit by git? That means you get the rename information at the top, and you can see the inline difs. Here, database.js wasn't picked up as a move, but rather as a delete and create, which means this whole file is marked as "new code" even though you only modified it (I believe)

src/shared/types/Season.ts
Lego6245: Fair fair. No need to refactor then.

src/shared/database.ts
Thaoden: Ah I read it as a comment on the first line.

Yeah, I see that, although I migrated it the same way via `git mv`. Not sure why it got marked as new file, maybe too much changed..?

src/shared/constants.ts
Thaoden: Referencing comment below: Make them `enum`s or `const array`s then?

Lego6245: Whoops, that must have gotten submitted automatically. No, revert your last commit and use the pattern above (appending `as const` to the end). The change you made will likely break everything as they're expecting arrays.

Thaoden: I just had a look in the transpiled js, looks like they are arrays there ;)

Thaoden:
Migrated two additional files: `Colors` and `CardTypes`

src/shared/database.ts
Thaoden: I _think_ this happened while stashing away some changes. Trying to rename the file again so that git takes up on the file change didn't help...

src/shared/database.ts
```diff
@@ -0,0 +1,264 @@
...
+    return this.metadata ? this.metadata.limited_ranked_events : {};
+  }
+
+  get standard_ranked_events(): { [id: number]: string } {
+    return this.metadata ? this.metadata.standard_ranked_events : {};
+  }
+
+  get single_match_events(): { [id: number]: string } {
+    return this.metadata ? this.metadata.single_match_events : {};
+  }
```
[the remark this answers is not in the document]
Thaoden: Regarding the file rename, that didn't work :/ But it broke Travis (at least something, eh? ;) )

src/window_background/data.ts
```diff
@@ -235,7 +235,7 @@ function matchIsLimited(match: MatchData): boolean {
   var eventId = match.eventId || match.InternalEventName;
 
   // The order of can matter.
-  if (database.limited_ranked_events.includes(eventId)) {
+  if (eventId && Object.keys(database.limited_ranked_events).includes(eventId)) {
```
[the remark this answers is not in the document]
Thaoden: (Related to https://github.com/Manuel-777/MTG-Arena-Tool/pull/684#discussion_r344427699)
`eventId` here is of type `string | undefined`. after the `if` it is of type `string` - but you cannot use a `string` type to access an array (I think - second opinion welcome) by index. That's why I changed the type of `database.limited_ranked_events` from `string[]` to a dictionary `id: string -> eventName: string`.

src/shared/database.ts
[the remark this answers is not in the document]
Thaoden: Regarding the type change, I commented [below](https://github.com/Manuel-777/MTG-Arena-Tool/pull/684#discussion_r344427727).

src/shared/database.ts
Thaoden: Regarding the file rename, that didn't work :/ But it broke Travis (at least something, eh? ;) )

src/window_background/data.ts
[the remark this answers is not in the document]
Thaoden: Meh I messed it up. 
`eventId` is not the index of a value in the array / dictionary, but rather the value of the array entry itself. I changed `database.limited_ranked_events` to be a `string[]`, along with the other `_events`.

src/shared/database.ts
lusbenjamin: Darn it. Git is smart enough to be stubbornly stupid in our case. Such a teenager.

Thaoden:
Closed in favor of #709.

PR #685, opened 2019-11-01
React card widgets

lusbenjamin:
### Motivation
This PR ports the easier widgets in `card-hover` into React function components:
  - `OwnershipStars`
  - `DraftRatings`

Since the original `card-hover` functions always set `parent.innerHTML= ""`, we can just let React control the entire parent and render the new Components there.


src/shared/DraftRatings.jsx
```diff
@@ -0,0 +1,46 @@
+import * as React from "react";
```
Lego6245: Is there a reason why these files aren't `.tsx`?

src/shared/DraftRatings.jsx
```diff
@@ -0,0 +1,46 @@
...
+import { DRAFT_RANKS } from "./constants";
+
+export default function DraftRatings(_props) {
+  const { card } = _props;
+  const rank = card.rank;
```
Lego6245: nit: destructuring assignment for these card properties

src/shared/DraftRatings.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <div className="rank_values_main_container">
+      <div className="rank_value_container">
+        Rank: {DRAFT_RANKS[Math.round(rank)]}
+      </div>
+      <div className="rank_value_container">Controversy: {rankControversy}</div>
+      {rankValues &&
+        rankValues.map((v, index) => {
+          const rv = 12 - index;
```
Lego6245: should probably be it's own component if you're gonna do all this logic here. also might be better to do a case/switch

Lego6245: also add a todo for when #684 goes in to go back and get the rank value properly typed here.

src/shared/OwnershipStars.jsx
```diff
@@ -0,0 +1,51 @@
...
+    return <div className={`inventory_card_infinity_${color}`} title={title} />;
+  }
+
+  title = `${owned || 0}/4 copies in collection`;
+  if (acquired) {
+    title += ` (${acquired} recent)`;
+  }
+  return (
+    <>
+      {[1, 2, 3, 4].map(i => {
```
Lego6245: nit: make this a const with a nice name

src/shared/OwnershipStars.jsx
```diff
@@ -0,0 +1,51 @@
+import * as React from "react";
+
+import pd from "./player-data";
+import { cardHasType } from "./card-types";
+
+export default function OwnershipStars(_props) {
+  const { card } = _props;
+  if (!card || !card.type) {
+    return <></>;
```
Lego6245: do we seriously have cases where this happens?

src/shared/OwnershipStars.jsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+  const isbasic = cardHasType(card, "Basic Land");
+  const owned = pd.cards.cards[card.id];
+  const acquired = pd.cardsNew[card.id];
+  let title = "";
+
+  if (isbasic) {
```
Lego6245: feels like we're rendering basically 3 different components here:
1) a no card component
2) a basic land component
3) a regular old component

since they're that different, maybe this would work better as 2 or 3 seperate components that are switched between based on the root if statements.

src/shared/DraftRatings.jsx
lusbenjamin: I am a coward. It looked like a proper upconversion to Typescript involved defining an interface for our vague `card` and/or `database` objects. Note my additional cowardly use of `_props` instead of `props` to dodge the linter rules.

src/shared/OwnershipStars.jsx
lusbenjamin: bug: classic fencepost error. should be `[0, 1, 2, 3]`

src/shared/OwnershipStars.jsx
lusbenjamin: This component began as an almost-literal translation of the utility function, which had to handle this case. I believe it was almost guaranteed to come up around the time when new sets/promo cards come out and the `database` object would come back with misses in some old code?

PR #686, opened 2019-11-01
React card tile

lusbenjamin:
### Motivation
This ports the `deck-drawer` module into React function components. It attempts to maintain backwards compatibility with a wrapper div that is safe for React to control without accidentally killing off DOM siblings. It has not yet been thoroughly tested for resulting visual regressions.


src/shared/CardTile.tsx
```diff
@@ -0,0 +1,443 @@
...
+        addCardHover(tileDiv, card, landOdds);
+      }
+    }
+  });
+
+  if (quantity === 0) return <></>;
+  if (!card) return <></>;
+
+  let quantityElement;
+  if (typeof quantity === "object") {
```
Manuel-777: I really dislike this check.. Maybe we can do something better here now we have react? or pass on a type?

src/shared/CardTile.tsx
```diff
@@ -0,0 +1,443 @@
...
+
+    default:
+      return "white";
+  }
+}
+
+function ArenaCardTile(props: CardTileProps): JSX.Element {
+  const {
+    card,
+    deck,
```
Manuel-777: Are these still passed by reference?

src/shared/CardTile.tsx
```diff
@@ -0,0 +1,443 @@
...
+  quantity: { quantity: string; odds: number } | number | string; // TODO clean this up?
+  showWildcards: boolean;
+  style: number;
+}
+
+function isNumber(n: number | string): boolean {
+  return !isNaN(parseFloat(n as string)) && isFinite(parseFloat(n as string));
+}
+
+function frameClassName(card: any): string {
```
Lego6245: Something about this makes me go hmm, is there a more efficient way to do this?

src/shared/CardTile.tsx
```diff
@@ -0,0 +1,443 @@
...
+
+  useEffect(() => {
+    if (card) {
+      const contId = `t${card.id + indent}`;
+      const tileDiv = byId(contId);
+      if (tileDiv) {
+        addCardHover(tileDiv, card, landOdds);
+      }
+    }
+  });
```
Lego6245: My intuition says this is right (you only need to bind this once per lifecycle), but I think you _may_ need to take a dependency on landOdds and card for this to properly re-execute.

src/shared/CardTile.tsx
```diff
@@ -0,0 +1,443 @@
...
+
+  return (
+    <div
+      className="card_tile_container_flat click-on"
+      id={contId}
+      data-grp-id={card.id}
+      data-id={indent}
+      data-quantity={quantity}
+      style={tileStyle}
+      onMouseEnter={(): void => setMouseHovering(true)}
```
Lego6245: nit: prefer this being wrapped in a `useCallback`.

src/shared/CardTile.tsx
```diff
@@ -0,0 +1,443 @@
...
+    landOdds,
+    quantity,
+    showWildcards
+  } = props;
+  const [isMouseHovering, setMouseHovering] = useState(false);
+
+  useEffect(() => {
+    if (card) {
+      const contId = `t${card.id + indent}`;
+      const tileDiv = byId(contId);
```
Lego6245: I see multiple usages of this function here. What you're most likely looking for is `React.useRef` instead of doing this jank where you're using query selector (clean up both usages with this). [See here for how to use.](https://reactjs.org/docs/hooks-reference.html#useref)

src/shared/CardTile.tsx
```diff
@@ -0,0 +1,443 @@
...
+      <div className="card_tile_quantity">
+        <span>{quantity}</span>
+      </div>
+    );
+  }
+
+  const costSymbols: JSX.Element[] = [];
+  const contId = `t${card.id + indent}`;
+  let wildcardElement = null;
+  if (card) {
```
Lego6245: This is prime real estate to move this into it's own component, no?

src/shared/CardTile.tsx
```diff
@@ -0,0 +1,443 @@
...
+        costSymbols.push(
+          <div
+            key={dfcCard.id + "_" + index}
+            className={"mana_s16 flex_end mana_" + cost}
+          />
+        );
+      });
+    }
+
+    if (showWildcards && deck) {
```
Lego6245: Ditto. Basically, if there's a place where you're making a component, you should consider moving it into it's own component instead of doing it as part of render. It's cleaner, more reusable, and keeps the render function leaner looking.

src/shared/CardTile.tsx
Manuel-777: I guess we can store a binary array with the string inside?
`tiles[1, 0] = "tile_c";
tiles[2, 0] = "tile_w;
...`
Then just access the string;
`return tiles[frame[0], frame[1] | 0];`

Manuel-777: Even better, we can use the color constants to declare the array;
```
tiles[COLORLESS, COLORLESS] = "tile_c";
tiles[WHITE, COLORLESS] = "tile_w;
tiles[BLUE, COLORLESS] = "tile_w;
```

src/shared/CardTile.tsx
lusbenjamin: yes

lusbenjamin:
@Manuel-777 I like the idea of using the new `mountReactComponent`, but keep in mind that this (and #685) are attempting to be more generic, cross-process utility components. `mountReactComponent` is in `window_main` instead of `shared` because it specifically handles clean up inside the main render process and not the overlay. @Lego6245 I don't know if that means we should pass some more plumbing around to ensure unmounting does what we expect in all cases.

Lego6245:
Yeah, that utility function only makes sense for `window_main`. We may need an analog for `window_overlay`, but I'm not sure if that's super high priority considering the usage pattern of overlay.

The reason why we have that helper method is to aid in disposing of the mounted components when we render new pages inside the main app. You're technically supposed to call `ReactDOM.unmountComponentAtNode` on containers hosting react components when you're modifying them (like, say, blanking the innerHTML of the page, like we do in `window_main`). If there's a similar analogue in `overlay` we should do it.

src/shared/CardTile.tsx
lusbenjamin: I'll take a shot at cleaning this up a bit.

src/shared/CardTile.tsx
```diff
@@ -0,0 +1,423 @@
...
+function frameClassName(card: any): string {
+  const frame = card ? card.frame.concat().sort() : [];
+  if (frame.length === 0) {
+    return "tile_c";
+  } else if (frame.length <= 2) {
+    const colorString = frame.map((i: number) => COLORS_ALL[i]).join("");
+    return "tile_" + colorString;
+  } else {
+    return "tile_multi";
+  }
```
lusbenjamin: arguably a less-efficient way to do this, but more legible and maintainable?

lusbenjamin: this also required renaming the CSS classes that had the colors in the "wrong" order

Manuel-777: I think its better and more efficient? the other way could end up in 10+ checks while this is just 1 check + 2 loops in average :man_shrugging:

PR #687, opened 2019-11-02
React overlay wip

lusbenjamin:
### Motivation
This PR converts the bulk of the display code in the overlay process into React function components. Where possible, it also reuses these new components to replace old shared utility functions.

This PR is stacked on #685 and #686.

#### React VS legacy boundary
This PR focuses on the `elements_wrapper` part of each overlay, which is where most of display widgets are:
```
      <div class="outer_wrapper">
          <div class="overlay_wrapper overlay_bg_image">
          </div>
      </div>
        <!-- now under React control -->
      </div>
      <div class="outer_wrapper top_nav_wrapper">
        <div class="flex_item overlay_icon click-on"></div>
        <div class="button settings click-on" style="margin: 0;"></div>
        <div class="button close click-on" style="margin-right: 4px;"></div>
      </div>
```
This deliberately "leaves behind" all of the IPC control and outer-chrome logic in the original `window_overlay_v3/overlay.jsx` for now.

The other remaining obvious legacy dependency is the hover code, which arguably got a little bit worse, especially the stuff related to grouping lands together.

TODO:
- [x] action log
- [x] deck type stats
- [x] deck mana curve chart
- [x] all the testing
- [x] move draft pick/pack state back out of react into legacy `overlay.jsx` (keep draft overlay in sync with most recent pick/pack)
- [x] merge in changes from #685 and #686 after they land


src/overlay/ActionLog.jsx
```diff
@@ -0,0 +1,68 @@
...
+    });
+  });
+
+  return (
+    <div key="overlay_decklist" className="overlay_decklist click-on">
+      {actionLog &&
+        actionLog.map((log, index) => {
+          const displayLog = { ...log };
+          displayLog.str = log.str.replace(
+            "<log-card",
```
Manuel-777: Not sure if there is a way to make these into react components wince htye are generated based ona string?

lusbenjamin: There isn't a way to serialize and deserialize individual instances of React components and pass them over IPC. I'm sure most careful designers would deny that there is even a proper use case for doing so, they might instead suggest we don't do any display logic in our `background` code and instead just pass well-typed and structured data through IPC and then use it as props for a full component in the display code.

lusbenjamin:
I am ready to promote this to a real PR as soon as we merge its dependents! 😸

Manuel-777:
> I am ready to promote this to a real PR as soon as we merge its dependents!

You got it!

Manuel-777:
Looks great to me, Interestingly it fixes one issue in `master` where hovering on the lands odds shows the missing card image below the land odds.

lusbenjamin:
converted into #692

PR #692, opened 2019-11-03
Almost-React Overlay

lusbenjamin:
### Motivation
This PR converts the bulk of the display code in the overlay process into React function components. It attempts to leave the code functionally identical, but there may be one or two places where it unintentionally fixes bugs. This is a port of the WIP draft #687.

#### Shared Code
  - `util.deckManaCurve` ==> converted into `DeckManaCurve`, now uses new React component
  - `util.deckTypesStats` ==> converted into `DeckTypesStats`, now uses new React component
  - `constants.DRAFT_RANKS` refactored to allows TS to infer its type (redundant with part of #684)

#### Other Non-Overlay Changes
  - `background/forceDeckUpdate` now includes the `sampleSize` in the response payload. This ensures that the overlay has the sample-size that matches the request when processing (instead of manually tracking it as state that could get out of sync in race conditions).

#### Overlay React VS legacy boundary
This PR focuses on the elements_wrapper part of each overlay, which is where most of display widgets are:
```
      <div class="outer_wrapper">
          <div class="overlay_wrapper overlay_bg_image">
          </div>
      </div>
        <!-- now under React control -->
      </div>
      <div class="outer_wrapper top_nav_wrapper">
        <div class="flex_item overlay_icon click-on"></div>
        <div class="button settings click-on" style="margin: 0;"></div>
        <div class="button close click-on" style="margin-right: 4px;"></div>
      </div>
```
This deliberately "leaves behind" the majority of the state management, "window" management, and IPC control in the original `window_overlay_v3/overlay.jsx` for now.

The other remaining obvious legacy dependency is the hover code, which can probably be refactored after this lands to clean up a lot of the unnecessary plumbing.

### DIFF
--- a/src/overlay/ActionLog.jsx
+++ b/src/overlay/ActionLog.jsx
@@ -1,0 +1,68 @@
+import React, { useEffect } from "react";
+import format from "date-fns/format";
+
+import db from "../shared/database";
+import { queryElements } from "../shared/dom-fns";
+import { addCardHover } from "../shared/card-hover";
+
+export default function ActionLog(_props) {
+  const { actionLog, index } = _props;
+  const initalTime = actionLog[0] ? new Date(actionLog[0].time) : new Date();
+
+  useEffect(() => {
+    const container = queryElements(
+      `#overlay_${index + 1} .overlay_decklist`
+    )[0];
+
+    const doscroll =
+      Math.round(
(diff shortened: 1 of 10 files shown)


lusbenjamin:
not ambitious enough!

PR #693, opened 2019-11-05
React Overlay (aka overlay_v3.14159)

lusbenjamin:
### Motivation
This PR converts the overlay process into React function components. It attempts to leave the code functionally identical, but there may be one or two places where it unintentionally fixes bugs. This is a port of the work in #687 and #692.

### Shared Code
  - `util.deckManaCurve` ==> converted into `DeckManaCurve`, now uses new React component
  - `util.deckTypesStats` ==> converted into `DeckTypesStats`, now uses new React component
  - `constants.DRAFT_RANKS` refactored to allows TS to infer its type (redundant with part of #684)
  - `card-hover` and `CardTile` updated:
    - `card-hover` handles legacy hover logic (main window only)
    - `CardTile` now supports React-based callback hover logic (overlay only at the moment)

### Other Non-Overlay Changes
  - `background/forceDeckUpdate` now includes the `sampleSize` in the response payload. This ensures that the overlay has the sample-size that matches the request when processing (instead of manually tracking it as state that could get out of sync in race conditions).
  - `gre-to-client-interpreter` now includes the `grpId` on each action log entry. This did not require any data model changes because it looks like @Manuel-777 had already started working in this direction?

### `npm install`
Friendly reminder that you have to do it. This PR updates our `interactjs` dependency, adds a dependency on `react-transition-group`, and adds several types-dependencies.

#### Introducing React Transition Group
We now have [React Transition Group](http://reactcommunity.org/react-transition-group/) to allow us to use CSS transitions on component lifecycle events. (@Manuel-777 and I played with this a bit during our earlier `create-react-app` experiment--this is a port of that work.)

#### New Overlay Modules and Files
- `index.html`, `index.css`, and a new `index.tsx`: render process entry point stuff. Mostly just copied over from the old `window_overlay_v3` gang.
- `OverlayController`: handles IPC traffic and all* overlay process state. Really wishes it had `redux`?
  - `player-data` still listens to `"set_player_data"` signals
- `CardDetailsWindowlet`: renders details about the currently hovered card
- `OverlayWindowlet`: renders one of the 5 overlay window widgets
- `MatchElements`: renders most of the contents of any of the match-mode overlays
- `DraftElements`: renders most of the contents of any of the draft-mode overlays
- `DeckList`: used in all modes, renders the subtitle, card tiles, and deck-related widgets.
- `ActionLog`, `SampleSizePanel`: smaller, self-descriptive match-mode sub-widgets

#### Remaining Technical Debt
- Using `player-data` in the React overlay starts to feel a little clunky, it probably wants to become a proper redux store?
- When we have more types available in `shared`, refactor common types out of `overlayUtil`
- the group lands hack is still a dirty hack


src/overlay/ActionLog.tsx
```diff
@@ -0,0 +1,71 @@
...
+          );
+
+          return (
+            <div className={"actionlog log_p" + log.seat} key={"log_" + index}>
+              <div className="actionlog_time" title={format(date, "HH:mm:ss")}>
+                {secondsPast + "s"}
+              </div>
+              <div
+                className="actionlog_text"
+                dangerouslySetInnerHTML={{ __html: log.str }}
```
lusbenjamin: if it's so dangerous, why does setting it feel so good?

src/overlay/OverlayController.tsx
```diff
@@ -0,0 +1,280 @@
...
+  useEditModeOnRef(editMode, hoverContainerRef);
+
+  const handleSaveOverlaysPosition = (): void => {
+    // Update each overlay with the new dimensions
+    const newOverlays = settings.overlays.map(
+      (overlay: OverlaySettingsData, index: number) => {
+        // TODO figure out a way to refactor this out
+        // some kind of useRef array?
+        const overlayDiv = byId("overlay_" + (index + 1));
+        let bounds = overlay.bounds;
```
lusbenjamin: I couldn't figure out a way to get rid of this last remaining `byId` logic. My first instinct is to use an array of `useRef`s, but the documentation specifically forbids using hooks inside of loops or if blocks or anything not the base level.

src/overlay/OverlayController.tsx
```diff
@@ -0,0 +1,280 @@
...
+  };
+
+  const handleClose = (
+    event: any,
+    arg: { action: any; index: number }
+  ): void => {
+    const bool = arg.action;
+    const index = arg.index;
+    // -1 to toggle, else set
+    const show = bool == -1 ? !settings.overlays[index].show : bool;
```
lusbenjamin: I think this can be `true | false | -1`, which seems to be a ridiculous API choice

src/overlay/OverlayController.tsx
```diff
@@ -0,0 +1,280 @@
...
+export default function OverlayController(): JSX.Element {
+  const [actionLog, setActionLog] = useState([] as LogData[]);
+  const [arenaState, setArenaState] = useState(ARENA_MODE_IDLE);
+  const [editMode, setEditMode] = useState(false);
+  const [match, setMatch] = useState(undefined as undefined | MatchData);
+  const [draft, setDraft] = useState(undefined as undefined | DraftData);
+  const [draftState, setDraftState] = useState({ packN: 0, pickN: 0 });
+  const [turnPriority, setTurnPriority] = useState(1);
+  const playerData = pd as any;
+  const [settings, setSettings] = useState(playerData.settings as SettingsData);
```
lusbenjamin: `OverlayController` definitely has `player-data` as implicit state. But I'm not fully confident that every change to player data is guaranteed to re-render what it needs to (e.g. the correct ownership star data).

(Note that `player-data` manages its own IPC traffic and state externally from `OverlayController`)

src/overlay/OverlayController.tsx
```diff
@@ -0,0 +1,280 @@
...
+      _priority == _we &&
+      settings.sound_priority
+    ) {
+      const { Howl, Howler } = require("howler");
+      const sound = new Howl({ src: ["../sounds/blip.mp3"] });
+      Howler.volume(settings.sound_priority_volume);
+      sound.play();
+    }
+    setTurnPriority(_priority);
+  };
```
lusbenjamin: this is a great candidate for some kind of memoization. Also can we do something more efficient with `Howl` here? I don't know any details about the dependency yet.

src/overlay/DeckList.tsx
```diff
@@ -0,0 +1,275 @@
...
+    subTitle,
+    settings,
+    tileStyle,
+    highlightCardId,
+    cardOdds,
+    setOddsCallback
+  } = props;
+  if (!deck) return <></>;
+
+  const deckClone = deck.clone();
```
lusbenjamin: we clone the deck to get away with all of the mutations we do (`removeDuplicates`, adding the grouped lands card, sorting)

src/overlay/OverlayWindowlet.tsx
```diff
@@ -0,0 +1,362 @@
...
+  index: number;
+  match?: MatchData;
+  settings: OverlaySettingsData;
+  tileStyle: number;
+  turnPriority: number;
+  setOddsCallback: (sampleSize: number) => void;
+  setDraftStateCallback: (state: DraftState) => void;
+}
+
+function OverlayElements(props: OverlayElementsProps): JSX.Element {
```
lusbenjamin: I know that `OverlayElements` is big enough to deserve its own file, but I really want to refactor it and `OverlayWindowlet`, so they live next to each other in this file for now.

src/overlay/OverlayWindowlet.tsx
```diff
@@ -0,0 +1,362 @@
...
+  //   xhr.open("HEAD", arg);
+  //   xhr.onload = function() {
+  //     if (xhr.status === 200) {
+  //       mainWrapper.style.backgroundImage = backgroundImage;
+  //     } else {
+  //       mainWrapper.style.backgroundImage = "";
+  //     }
+  //   };
+  //   xhr.send();
+  // }, [backgroundImage]);
```
lusbenjamin: Do we need anything like this in React land?

src/overlay/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const wrap = createDiv([]);
+  ReactDOM.render(<OverlayController />, wrap);
+  document.body.appendChild(wrap);
+  setTimeout(() => {
+    new TransparencyMouseFix({
+      log: false,
+      fixPointerEvents: "auto"
+    });
+  }, 1000);
+});
```
lusbenjamin: @Lego6245 any comments on the basic wiring of the root components here? We can "do the right thing" now that almost everything in the render process is React.

src/shared/constants.js
```diff
@@ -656,21 +656,22 @@ export const ARENA_MODE_IDLE = 0;
...
+  "C-",
+  "C",
+  "C+",
+  "B-",
+  "B",
+  "B+",
+  "A-",
+  "A",
+  "A+"
+];
```
lusbenjamin: redundant with a piece of #684

Thaoden: I can fix that conflict when it's time to merge 684. No worries.

src/overlay/OverlayWindowlet.tsx
lusbenjamin: Refactored!

lusbenjamin:
TODO:
- [ ] re-implement `interactjs` restriction for dragging and resizing

lusbenjamin:
Knocked off some more technical debt:
- ~~The biggest remaining obvious legacy dependency on `shared` is the hover code, which can probably be refactored after this lands to clean up a lot of the unnecessary plumbing.~~
- ~~we may want to more liberally apply `useMemo` and the optional arguments to `useCallback` and `useEffect` to optimize performance.~~

...does this mean the _ENTIRE_ overlay is now React?

lusbenjamin:
As part of getting to 100% react-rendered and controlled, this PR now changes the action log widget hover behavior:
react-action-log-hover

lusbenjamin:

PR #698, opened 2019-11-09
Update it all

Manuel-777:
Just update all packages into their latest versions.

`electron-builder` is already updated on `master`, and `èlectron-updater` needs to be tested (but this can only be done in master afaik?).


src/window_background/mtga-log.js
```diff
@@ -43,6 +43,6 @@ export async function readSegment(path, start, length) {
     const { bytesRead } = await fsPromises.read(fd, buffer, 0, length, start);
     return buffer.toString("utf-8", 0, bytesRead);
   } finally {
-    fs.close(fd);
```
lusbenjamin: I wonder if this change is necessary because of the TypeScript upgrade to `3.7`. This part of their release announcement may be relevant:
image

lusbenjamin:

PR #694, opened 2019-11-06
Reactify top nav

Manuel-777:
This PR streamlines the inner workings of the top navigation bar, simplifies the process of opening things like tabs or settings and creates a set of React components to replace the HTML5 mess we had going on there.

Extra goodies;
- Accidentally fixed some of the mess `renderer.js` was.
- Fixed some issues arising when moving trough tabs.
- Isolated display of top navigation from its control.

First run at tsx, reviewers beware!

PR #699, opened 2019-11-09
2.13.0

lusbenjamin:
This is our first experiment with an integration-branch! Please everyone feel free to edit and contribute to this.

### Also Included Feature PRs:
This list includes PRs that were part of the orginal branch and so will not get a pretty-looking squash-merge.
- #669
- #694
- #698
- #693

### TODO:
- [x] fix card hover on Economy page
- [x] merge in ~~#684~~ #709


Manuel-777:
One issue I noticed with the overlays; When settings are updated (click on any setting, move trough tabs) some times the overlays are displayed even when there is no active overlay. Causing the window to be irresponsive (overlay covers the screen and is invisible).

To see this I enabled DevTools on the overlays and selected `body`. Then went into settings page, clicked some section and see how body now covers everything with no overlay elements. Enabling an overlay does not allow clicking trough the transparent areas, the only fix is disabling the overlay so settings are updated and the overlay is minimized.

Thaoden:
Leave #684 out of this for a while, I'll try to port it onto the current dev as a new PR.

Manuel-777:
Also, it seems Sort filter in decks page (and other pages) is not working properly since #665 @Lego6245 
Tried fixing it but only broke it further :man_facepalming:

Lego6245:
See #705 for fixes.

src/shared/dom-fns.ts
```diff
@@ -0,0 +1,54 @@
...
+  return parentNode.querySelectorAll(selectors);
+}
+
+export function queryElementsByClass(selectors:string, parentNode = document):any {
+  return parentNode.getElementsByClassName(selectors);
+}
+
+type returnableElements = HTMLDivElement | HTMLImageElement | HTMLInputElement | HTMLLabelElement | HTMLSpanElement;
+
+function createElement(
```
Thaoden: You can constrain the return type to `returnableElements` here, as you return the element passed in. nit: Maybe name the function differently, as it does not _create_ but rather _enrich_(?) the element.

Thaoden: Alternatively, pull the `document.createElement()` from the callers into this function and just pass in the html item string.

Thaoden: Should / can `attrs` be typed?

PR #700, opened 2019-11-10
fix hovers

Manuel-777:
Fixes the hovers on economy page.

### DIFF
--- a/src/window_main/components/economy/EconomyRow.tsx
+++ b/src/window_main/components/economy/EconomyRow.tsx
@@ -12,7 +12,10 @@ import {
   openScryfallCard,
   getCardArtCrop,
 } from "../../../shared/util";
-import React from "react";
+
+import { addCardHover } from "../../../shared/card-hover";
+
+import React, { useState } from "react";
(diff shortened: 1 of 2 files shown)


src/window_main/components/economy/EconomyRow.tsx
```diff
@@ -248,16 +251,22 @@ interface InventoryCardProps {
...
     const lookupCard = card.dfc == "SplitHalf" ? db.card(card.dfcId) : card;
     openScryfallCard(lookupCard);
   }, [card]);
   // inventoryCard.style.width = "39px";
 
+  React.useEffect(() => {
+    if (inventoryCardRef) {
+      addCardHover(inventoryCardRef.current, card);
+    }
+  });
```
lusbenjamin: nit: I _think_ that theoretically the last line could be `}, [card]);` to ensure that the effect only gets re-run if `card` changes. I'm not sure how often in practice we do anything other than blow the whole thing away?

Manuel-777: Yeah, the card is pretty much static until its destroyed. I thought about adding a return to the useEffect to blow out the event listener created but GC takes care of that already.

PR #703, opened 2019-11-10
Fix display of aetherized and added cards

Thaoden:
Quick fix for the economy display, trying to display aetherized cards or cards added to collection.

Sorry for the whitespace change noise, that was a VS Code formatter. When reviewing, look for the gear icon and select the checkbox on "Ignore whitespace changes", click "Apply & Reload" and it should be more clear what happened. This setting resets on the next review.

PR #705, opened 2019-11-10
Targeted bugfixes for develop branch

Lego6245:
1) Fixes problem where either the addedCards or aetherized cards were null.

2) Fixes sort not working in explore page (we need to refactor it so that it can sort without executing a new search every time)

### DIFF
--- a/src/window_main/FilterPanel.tsx
+++ b/src/window_main/FilterPanel.tsx
@@ -127,7 +127,7 @@ class FilterPanel {
               <ManaFilter prefixId={this.prefixId} filterKey={"colors"} filters={this.filters} onFilterChanged={this.filterCallback("colors")} />
             )}
         </div>)}
-        {(this.archs.length || this.showOppManaFilter) && (
+        {(this.archs.length || this.showOppManaFilter || this.showSortOption) && (
           <div className={"columnC"}>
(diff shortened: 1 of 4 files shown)


Thaoden:
The addedCards / aetherizedCards is adressed in #703 already.

Lego6245:
#703’s change feels a little messier to me. I’m willing to drop the fix from this changeset though

src/window_main/FilterPanel.tsx
```diff
@@ -127,7 +127,7 @@ class FilterPanel {
               <ManaFilter prefixId={this.prefixId} filterKey={"colors"} filters={this.filters} onFilterChanged={this.filterCallback("colors")} />
             )}
         </div>)}
-        {(this.archs.length || this.showOppManaFilter) && (
+        {(this.archs.length || this.showOppManaFilter || this.showSortOption) && (
           <div className={"columnC"}>
             {this.archs.length && <WrappedReactSelect className={this.prefixId + "_query_optag filter_panel_select_margin spacer180"} options={this.archs} current={this.filters.arch} callback={this.filterCallback('arch')} optionFormatter={(tag: string) => <TagOption tag={tag} showCount={true} archCounts={this.archCounts} />} />}
```
Manuel-777: I believe this should be an if? 
`this.archs.length ? <`

The way it is now adds a "0" before the sort selector in the DOM.

Lego6245:
Good catch!

PR #709, opened 2019-11-11
Shared code TypeScript migrations (Take Two)

Thaoden:
Let's try this again!

New version of #684.


Thaoden:
I seriously don't know why Git is having such problems registering the filename change as what it is...

Thaoden:
Wheee!!!

PR #711, opened 2019-11-12
Just More Types

lusbenjamin:
This is mostly me cleaning up a bunch of `any` types that I left lying around. ~~Should have no impact on the transpiled output.~~ Should be logically equivalent to the previous transpiled output. (But `null !== undefined`.)


src/overlay/CardDetailsWindowlet.tsx
```diff
@@ -35,7 +36,7 @@ const SCALAR = 0.71808510638; // ???
 
 export interface CardDetailsWindowletProps {
   arenaState: number;
-  card?: any;
+  card?: DbCardData | any;
```
Thaoden: This still means that `card?` is of type `any`. Maybe this should be `card? : DbCardData`?

src/shared/CardTile.tsx
```diff
@@ -15,16 +15,17 @@ import {
   openScryfallCard
 } from "./util";
 import { addCardHover } from "./card-hover";
+import { DbCardData } from "./types/Metadata";
 
 export interface CardTileProps {
-  card: any;
-  deck: Deck | null;
-  dfcCard: any;
+  card: DbCardData | any;
```
Thaoden: Same as above - still of type `any`.

src/shared/CardTile.tsx
```diff
@@ -174,7 +175,7 @@ function ArenaCardTile(props: CardTileProps): JSX.Element {
   }, [setHoverCardCallback]);
   const handleMouseLeave = useCallback((): void => {
     setMouseHovering(false);
-    setHoverCardCallback && setHoverCardCallback(null);
+    setHoverCardCallback && setHoverCardCallback();
```
Thaoden: This looks funny...

src/shared/CardTile.tsx
```diff
@@ -315,7 +316,7 @@ function FlatCardTile(props: CardTileProps): JSX.Element {
...
```
Thaoden: As does this.

src/overlay/CardDetailsWindowlet.tsx
lusbenjamin: this might be worth a code comment. Every place in the diff that you see `DbCardData | any` is one of the locations that we have to handle the "Grouped Lands Hack", which means we have to expect either a well formed card or a card shoved inside a card (no data type for that second one).

src/shared/CardTile.tsx
lusbenjamin: yep. I usually prefer `null`.

src/overlay/CardDetailsWindowlet.tsx
Thaoden: ... yet ;) (I hope)

lusbenjamin: nah, we definitely just fix the hack instead of ossifying it into a formal set of types

src/shared/CardTile.tsx
Thaoden: What would break if you'd mark `setHoverCardCallback` as non-nullable?  You wouldn't have to check it but can call it directly instead. But I guess that's a matter of taste...

lusbenjamin: We would have to update the legacy code that expects to use `CardTile` with `addCardHover` instead of a proper React-style callback passed through props.

I'm also not sure that we want all uses of `CardTile` to require specifying a hover handler, so it does seem to fit the bill of an optional prop.

Thaoden: Fair enough. Still looks funny ;)

Thaoden:
Looks good to me!

Issue #602, opened 2019-09-30
Enabling edit mode is unintuitive


dpdearing:
I'm running in a Windows VM through Parallels and I can no longer drag overlays around the desktop to reposition them.

This has been happening to me since around 2.9.3 or so (maybe earlier), but was fine previously.  I tried renaming my config file like the FAQ said, but no change except now they are in the default location and I still can't move them.

AnnanFay:
In order to move overlays around you need to switch to edit mode. Either using the shortcut `Alt + Shift + E` or by clicking on the weird little coloured circle thing on an existing overlay.

![image](https://user-images.githubusercontent.com/766048/65928939-c438e100-e3f7-11e9-8811-4dc8aae663cb.png)

dpdearing:
Thanks!  There have been a lot of little tweaks like this that aren't intuitive from the UI.

AnnanFay:
> There have been a lot of little tweaks like this that aren't intuitive from the UI.

Yeah. I think this is still a valid bug. I'm going to reopen this for now and label as User Experience (UX). It's come up multiple times in the discord server as well.

Potential fixes for this are:

1. Hover over tooltip on edit button - text something like "Toggle Edit Mode (Ctrl + Shift + E)"
2. Cross out the entire screen as well as dim it. This is to make it clearer that you cannot do anything during edit mode - it's a temporary mode.
3. Add "Toggle Edit Mode" to the right click taskbar menu
![image](https://user-images.githubusercontent.com/766048/65999373-2bf13980-e495-11e9-904c-77e8da4ace74.png)
.
4. Add explicit "Edit Mode" button to the overlay settings screen. 
![image](https://user-images.githubusercontent.com/766048/65999335-18de6980-e495-11e9-92b9-ebce2afb551b.png)

dpdearing:
I think the issue for me was that it wasn't even clear to me that an "Edit Mode" existed.  And normal GUI behavior is to just drag+drop a panel, so I didn't event think to look for a way to enable it.  Options 3 and 4 at least help with the issue of not knowing Edit Mode existed.

That said, once I got there it was really hard for me to use Edit Mode mid-game (because that's when the overlays show up) to adjust things.  All while my opponent is waiting for me to act and spamming "Your Go" as I fumble to move things and get out of Edit Mode with my timer burning.

Why does "Edit Mode" need to be a thing versus the old (and normal) drag behavior?  That (to me, anyway) was the source of the problem.

[...]

PR #717, opened 2019-11-13
Discoverable Edit Mode

lusbenjamin:
### Motivation
We have received several instances of feedback that the current "Edit Mode" feature of the Overlays is hard to find and/or understand, see #602.

This PR makes several changes to attempt to make "Edit Mode" more discoverable:
  - adds an "Edit Overlay Positions" button to the settings page with hover text
  - reorganizes the common settings on the "Settings > Overlay" page
  - standardizes the display text on "Edit Overlay Positions" in the context menu/shortcuts page
  - ensures that all enabled overlays are visible to move/resize during arena idle mode

### Demo
image

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -44,6 +44,7 @@ let mainLoaded = false;
 let backLoaded = false;
 let overlayLoaded = false;
 let arenaState = ARENA_MODE_IDLE;
+let isEditMode = false;
 
 const singleLock = app.requestSingleInstanceLock();
 
@@ -225,6 +226,12 @@ function startApp() {
           // set progress to >1 for indeterminate time
           mainWindow.setProgressBar(arg.progress);
         }
+        break;
+
+      case "edit":
+        isEditMode = !isEditMode;
+        updateOverlayVisibility();
+        overlay.webContents.send("edit");
         break;
 
       case "renderer_window_minimize":
@@ -447,7 +454,8 @@ function getOverlayVisible(settings) {
     (OVERLAY_DRAFT_MODES.includes(settings.mode) &&
       arenaState === ARENA_MODE_DRAFT) ||
     (!OVERLAY_DRAFT_MODES.includes(settings.mode) &&
-      arenaState === ARENA_MODE_MATCH);
+      arenaState === ARENA_MODE_MATCH) ||
+    (isEditMode && arenaState === ARENA_MODE_IDLE);
 
   return settings.show && (currentModeApplies || settings.show_always);
 }
@@ -618,7 +626,7 @@ function createMainWindow() {
       }
     },
     {
-      label: "Edit Mode",
+      label: "Edit Overlay Positions",
       click: () => {
         overlay.webContents.send("edit");
       }
--- a/src/main.js
+++ b/src/main.js
@@ -228,10 +228,8 @@ function startApp() {
         }
         break;
 
-      case "edit":
-        isEditMode = !isEditMode;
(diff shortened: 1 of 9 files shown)


src/overlay/OverlayController.tsx
```diff
@@ -191,12 +191,8 @@ export default function OverlayController(): JSX.Element {
   );
 
   const handleSettingsUpdated = useCallback((): void => {
-    // mid-match Arena updates can make edit-mode difficult
-    // temporarily allow the overlays to go stale during editing
-    // (should be okay since ending edit-mode causes a refresh)
-    if (editMode) return;
     setSettings({ ...playerData.settings });
-  }, [editMode]);
```
lusbenjamin: this hack should no longer be necessary in React-land

lusbenjamin:
currently a bit buggy: issues with edit mode state getting out of sync between processes and the overlay window not having correct visiblilty

lusbenjamin:
Bugs resolved! IPC signals are cleaner and state is synchronized correctly between processes.

lusbenjamin:
image

src/overlay/OverlayController.tsx
AnnanFay: One interesting option is to add a crossed out layer to partially hide content. This both signals the content is no longer valid and clearly indicates editing is enabled.

src/main.js
```diff
@@ -447,7 +454,8 @@ function getOverlayVisible(settings) {
     (OVERLAY_DRAFT_MODES.includes(settings.mode) &&
```
AnnanFay: This could do with a comment explaining what it's trying to do.

src/overlay/OverlayWindowlet.tsx
```diff
@@ -93,7 +94,8 @@ export default function OverlayWindowlet(
     (OVERLAY_DRAFT_MODES.includes(overlaySettings.mode) &&
       arenaState === ARENA_MODE_DRAFT) ||
     (!OVERLAY_DRAFT_MODES.includes(overlaySettings.mode) &&
-      arenaState === ARENA_MODE_MATCH);
+      arenaState === ARENA_MODE_MATCH) ||
+      (editMode && arenaState === ARENA_MODE_IDLE);
```
AnnanFay: Should this be `isEditMode` to have consistent variable names?

src/window_main/settings.js
```diff
@@ -465,6 +466,45 @@ function appendArenaData(section) {
 function appendOverlay(section) {
   section.appendChild(createDiv(["settings_title"], "Overlays"));
 
+  const displayControls = createDiv([]);
+  displayControls.style.display = "inline-flex";
```
AnnanFay: Can this be moved to the CSS file?

AnnanFay: Do we have guidelines between when to apply inline styles with JS versus creating new CSS classes? Using CSS seems like it will be universally better for themes later on.

src/overlay/OverlayController.tsx
lusbenjamin: This pointed out to me that we have a regression: the gray "blocking" div no longer blocks user input for this build. Which behavior do we want and why?

src/overlay/OverlayWindowlet.tsx
lusbenjamin: renaming `isEditMode` ==> `editMode` for consistency, doing the other one involves touching many files

PR #718, opened 2019-11-13
super special feature

lusbenjamin:
image
👺


lusbenjamin:
closing due to potential performance concerns, we may revisit this on 2020/04/01...

PR #720, opened 2019-11-14
More Migrations to TS

Thaoden:
Just some more types that fell off the table in the latest PR.

*Please* have a look at line 721 of `constants.ts`, I had to include `Lands` to fix some downstream type constraints.

And I am really sorry that the file rename wasn't picked up again. Maybe can someone explain to me how to do it properly? Aside from that, it's basically the same changes as in #684.


lusbenjamin:
image

src/shared/constants.ts
```diff
@@ -0,0 +1,803 @@
...
+  "Creatures",
+  "Lands",
+  "Instants",
+  "Sorceries",
+  "Enchantments",
+  "Artifacts",
+  "Planeswalkers"
+] as const;
+
+export const CARD_RARITIES = ["Common", "Uncommon", "Rare", "Mythic", "Land"] as const;
```
lusbenjamin: this is the line to consider carefully re: "Lands"

On a related note, does the change in case from lower to capital-case have an impact?

Thaoden: I had to change it so that it matches the `Rarities` enum from `Metadata.ts`, both are compared in `CardTile.tsx`, line 141 and line 284.

Thaoden: An alternative would be to check `card.rarity` beforehand if it equals to `Lands`, if it does, skip this Card Tile, if not, cast to `CARD_RARITIES` and proceed as usual.

lusbenjamin: I am particularly worried about regressions here:
image

Thaoden: Yeah, I've seen that too, though only after opening the PR. I'll revert the changes here and come up with a different solution.

Manuel-777: In abscraction "Lands" *is* a rarity, just not a useful one to have around most of the times. I think I had it listed at some point then removed it because it made a mess of use cases where we sometimes need it and sometimes not. Most of the times we didnt.

Manuel-777: If it becomes relevant we can totally have two different lists of rarities, since the most common use case to exclude lands is when manipulating wildcards, I belive we could have `WILDCARD:RARITIES` and `CARD_RARITIES`.
Now the question is if it is relevant enough?

Thaoden: What really bugs me though is that we have the rarities (almost, not quite) duplicated in `CARD_RARITIES` and `type Rarity`. That's the reason for the type conflicts here. I think it is cleaner to make them identical and check at the appropriate places if the rarity currently handled is a `Land` and then act accordingly (like, skipping it).

Thaoden:
Multiple things were adressed:
* Don't duplicate the possible card rarities. Rather use the constants in `constants.ts` to define the string union type in `Metadata.ts`.
* This meant that for the most part (all of them?) of the loops over `CARD_RARITIES`, `Land` was explicitly excluded...
* ... and the rarity was lowercased as to not break existing code.
* Some refactors along the way, eg. variables renames and a single `switch` instead of extended `if`s in `collection.js`. Most notably though a refactor in `CardTile.tsx` to remove duplicated code.

Thaoden:
Added the migration from `colors.js` to `colors.ts`. GitHub showed this as a rename until the `c02e122 - Refactorings` commit, then it gave up.

src/shared/colors.ts
```diff
@@ -0,0 +1,154 @@
...
+  MULTI,
+  COLORLESS
+} from "./constants";
+
+class Colors {
+  private w: number;
+  private u: number;
+  private b: number;
+  private r: number;
+  private g: number;
```
lusbenjamin: Does any non-TS code depend on getting or setting these fields as public?

lusbenjamin: @Manuel-777 says we should be g2g

Thaoden: I had a look at all the uses of `color.js` in every file - I didn't see any, so I decided it would be okay :)

PR #724, opened 2019-11-15
bugfix metadata js error

lusbenjamin:
Solves a JS error introduced during some of the recent typing:
https://discordapp.com/channels/463844727654187020/506793351786266624/644641855119687690
image


Thaoden:
I would have liked it more to keep the private prop private and expose it via a getter. No need for some external class to change my internal data. But that can be fixed once all has been migrated to sunny TS land where the grass is green :)

PR #725, opened 2019-11-16
Deck and cards list to ts

Manuel-777:
This is a tricky one, and might be full of things to reconsider in a TS world.


Thaoden:
I have the same one in the pipeline :D though I suggest waiting for #720 and using some of the "new" types there.

src/shared/cards-list.ts
```diff
@@ -1,15 +1,39 @@
 import _ from "lodash";
```
Thaoden: this file should probably be named `cardsList.ts`

src/shared/cards-list.ts
```diff
@@ -1,15 +1,39 @@
...
 import db from "./database.js";
 import Colors from "./colors.js";
+import { v2cardsList, v3cardsList, anyCardsList } from "./types/Deck";
+import { DbCardData } from "./types/Metadata.js";
+
+interface cardTypesCount {
```
Thaoden: I'd PascalCase this (and the next).

src/shared/cards-list.ts
```diff
@@ -35,14 +59,14 @@ class CardsList {
...
+  get():v2cardsList {
     return this._list;
   }
 
   /**
    * Adds a card to the list
    **/
-  add(grpId, quantity = 1, merge = false) {
+  add(grpId:number, quantity:number = 1, merge:boolean = false) {
     if (typeof quantity !== "number") {
```
Thaoden: this check has become obsolete with the typings.

src/shared/cards-list.ts
```diff
@@ -66,24 +90,22 @@ class CardsList {
   /**
    * Removes a card from the list.
    **/
-  remove(grpId, quantity = 1, byName = false) {
+  remove(grpId:number, quantity:number = 1, byName:boolean = false) {
     if (typeof quantity !== "number") {
```
Thaoden: As has this one.

src/shared/cards-list.ts
```diff
@@ -66,24 +90,22 @@ class CardsList {
...
       throw new Error("quantity must be a number");
     }
     if (byName) {
-      let removed = 0;
-      let cardToFind = db.card(grpId);
+      let cardToFind = db.card(grpId) as DbCardData;
       this._list.forEach(function(card) {
-        let cardInList = db.card(card.id);
+        let cardInList = db.card(card.id) as DbCardData;
```
Thaoden: this is difficult. you can't be sure at runtime you really have a card. better check with `if()`

src/shared/cards-list.ts
```diff
@@ -244,19 +266,10 @@ class CardsList {
     return colors;
   }
 
-  /**
-   * Inserts a new propierty to each card in the list.
-   **/
-  addProperty(_prop, _default = 0) {
```
Thaoden: this one is used at least twice in different js files, always with the "_prop = chance`. this property should be added to `this` and a `addChance(_default)` added.

src/shared/cards-list.ts
```diff
@@ -1,15 +1,39 @@
 import _ from "lodash";
 import db from "./database.js";
 import Colors from "./colors.js";
+import { v2cardsList, v3cardsList, anyCardsList } from "./types/Deck";
```
Thaoden: are ' v3cardsList` and `anyCardsList` used anywhere?

src/shared/cards-list.ts
Manuel-777: You might have noticed this is a common check I do in this PR. I had to because the return type of `db.card()` is `DbCardData | false`, thus checking against the properties correctly warns the property does not apply for `false`.

Thaoden: `as` is not a type check though, afaik? It's a cast, you're telling the compiler "I _know_ it's of this type".

src/shared/deck.ts
```diff
@@ -1,5 +1,5 @@
 import db from "./database.js";
-import CardsList from "./cards-list.js";
+import CardsList from "./cardsList.js";
```
Thaoden: nit: `ts`

src/shared/cardsList.ts
```diff
@@ -174,37 +178,40 @@ class CardsList {
   /**
    * Creates an object containing the colors distribution of the list.
    **/
-  getColorsAmmounts():colorsCount {
+  getColorsAmounts():colorsCount {
```
Thaoden: The name changed, are we sure it's changed in all callers?

src/shared/constants.ts
```diff
@@ -718,7 +718,7 @@ export const CARD_TYPES = [
   "Planeswalkers"
 ] as const;
 
-export const CARD_RARITIES = ["Common", "Uncommon", "Rare", "Mythic", "Land"] as const;
+export const CARD_RARITIES = ["common", "uncommon", "rare", "mythic", "land"] as const;
```
Thaoden: making these lowercase could break `DbCardData`, should be tested.

Manuel-777: not making them lowercase breaks on https://github.com/Manuel-777/MTG-Arena-Tool/blob/16ea037b1ee77bd7d0c2df3bebb7c7b362c6ed7a/src/shared/deck.ts#L144

Not sure how to fix that without creating a new rarity in lowercase?

src/shared/cardsList.ts
Manuel-777: interestingly I cant find any usage of this functions in the code :man_shrugging:

src/shared/constants.ts
Thaoden: Migrate the database data to use lowercase rarities *duck & run

src/shared/cardsList.ts
Thaoden: I had the same feeling. Remove it, then?

src/window_background/forceDeckUpdate.ts
```diff
@@ -15,20 +52,21 @@ const forceDeckUpdate = function(removeUsed = true) {
   globals.currentMatch.playerCardsLeft = globals.currentMatch.player.deck.clone();
 
   if (globals.debugLog || !globals.firstPass) {
-    globals.currentMatch.playerCardsLeft.mainboard.get().forEach(card => {
+    globals.currentMatch.playerCardsLeft.mainboard.get().forEach((card: any) => {
```
Thaoden: Can this be typed as `CardObject`?

src/shared/cardsList.ts
```diff
@@ -31,13 +31,18 @@ class CardsList {
    * If an array of IDs is given it sets each quantity to the number of adjacent
    * repetitions
    **/
-  private list: v2cardsList;
+  public list: v2cardsList;
```
Thaoden: I'd change this back to `private` and use a getter if this need to be accessed from outside. No external entity should be able to directly modify something internal to this class.

src/shared/cardsList.ts
```diff
@@ -306,7 +301,10 @@ class CardsList {
 
     this.list.forEach(function (card) {
       let cardObj = db.card(card.id);
-      let found = newList.find((c: CardObject) => { let dbCard = db.card(c.id); return dbCard && cardObj && dbCard.name === (cardObj as DbCardData).name });
+      let found = newList.find((c: CardObject) => {
+        let dbCard = db.card(c.id);
+        return dbCard && cardObj && dbCard.name === (cardObj as DbCardData).name
```
Thaoden: nit: missing semicolon?

src/shared/deck.ts
```diff
@@ -25,19 +25,21 @@ interface internalDeck {
...
-  private _colors: Colors;
-  private tags: string[];
-  private custom: boolean;
+  public id: string;
+  public lastUpdated: string;
+  public tile: number;
+  public _colors: Colors;
+  public tags: string[];
+  public custom: boolean;
+  public archetype: string;
```
Thaoden: Why have all these changed from `private` to `public`?
Btw, I think `public` is the default, so you could omit the access modifier.

src/window_background/data.ts
```diff
@@ -7,6 +7,8 @@ import playerData from '../shared/player-data.js';
 import { DEFAULT_TILE } from '../shared/constants';
```
Thaoden: This file is quite long. Maybe extract the interfaces into their own files?

src/window_background/labels.js
```diff
@@ -266,20 +266,24 @@ function select_deck(arg) {
...
-  return JSON.parse(JSON.stringify(deck), (key, value) => {
-    if (key === "mainDeck" || key === "sideboard") {
-      let ret = [];
-      for (let i = 0; i < value.length; i += 2) {
-        if (value[i + 1] > 0) {
-          ret.push({ id: value[i], quantity: value[i + 1] });
-        }
-      }
-      return ret;
+function convertV3ToV2(v3List) {
```
Thaoden: Isn't this the same as calling the `CardsList` constructor with the `v3List` param?

src/shared/cardsList.ts
Manuel-777: I went dumb for a second and forgot about getters!

src/shared/deck.ts
Manuel-777: There are places in which we access `commandZoneGRPIds` directly, so I thought adding the rest as public was okay as well.

src/window_background/labels.js
Manuel-777: This only modifies an array, while `CardsList` creates a class for it. I use this in cases where we need to preprocess the data before jamming it in a `Deck` or `CardsList`.
We can totally use a `CardsList` to convert but it seems overkill to do that and then dispatch it? maybe we can look at this in depth when we convert labels to ts.

src/shared/deck.ts
Thaoden: Accessing directly should not be necessary / should be prevented. What about introducing a getter and / or setter that exposes `commandZoneGRPIds` to externals?

src/window_background/labels.js
Thaoden: I don't know enough of JS memory management, but revisiting surely is not a bad idea :)

src/shared/deck.ts
```diff
@@ -37,22 +69,22 @@ class Deck {
    * returns the colors of this deck, or creates a new colors object
    * if not defined yet.
    **/
-  get colors() {
+  get colors():Colors {
     return this._colors || this.getColors();
   }
```
Thaoden: `this._colors` is now always instantiated (line 57), so the check won't fail anymore and won't ever call `this.getColors()`. Can this directly return `this.getColors()`?

Manuel-777: You are right.. but this is wrong now I see it, the first call to get the colors should get the actual colors, so maybe the `constructor` instantiator should call the getter instead if creating an ampty Color

src/shared/deck.ts
```diff
@@ -70,7 +70,7 @@ class Deck {
    * if not defined yet.
    **/
   get colors():Colors {
-    return this._colors || this.getColors();
+    return this._colors;
```
Thaoden: I'm pretty sure this will produce a stack overflow on first call to the constructor ;)

Thaoden: Is this needed any longer? or just replace all calls to `deck.color` by `deck.getColors()` (or the other way round).

Manuel-777: `getColors()` actually loops trough all cards and re-creates `this._colors`, `colors()` just access the stored value (is better in most cases).

src/window_background/labels.js
```diff
@@ -508,7 +508,7 @@ export function onLabelClientToMatchServiceMessageTypeClientToGREMessage(
     let newDeck = globals.currentMatch.player.deck.clone();
     newDeck.mainboard = tempMain;
     newDeck.sideboard = tempSide;
-    newDeck.getColors();
+    newDeck.colors();
```
Thaoden: probably a typo?

PR #726, opened 2019-11-17
Migrate card-types to TS

Thaoden:
Just a small migration.


src/shared/OwnershipStars.tsx
```diff
@@ -65,7 +65,7 @@ export default function OwnershipStars(props: { card: DbCardData }) {
   if (!card || !card.type) {
     return <></>;
   }
-  const isbasic = cardHasType(card, "Basic Land");
+  const isbasic = cardHasType(card, "Land");
```
Manuel-777: This is wrong?

Thaoden: I think it was a bug before. I didn't change the types, so `cardHasType(card, "Basic Land")` _always_ returned false.

lusbenjamin: Should be an easy empirical test: you need to see an infinity symbol over all of your basic lands in the "Collection" page, but normal 4 star pips over land that requires playsets.

Thaoden: I see the infinity symbol alright, but I think I don't have those other lands for comparison.

Thaoden: It was a bug in my change. `card.type` is a string, so can contain anything and is not constrained to types.

src/window_main/renderer-util.js
```diff
@@ -220,6 +220,7 @@ function drawDeck(div, deck, showWildcards = false) {
         case "Artifact":
           return "Artifacts";
         case "Land":
+        case "Basic Land":
```
Manuel-777: I thought this handled partial matches? (thus why there was only "Land") What would happen with things like "Artifact Creature" or "Instant Sorcery" like some split cards?

Thaoden: I had to add "Basic Land" because I added that one to the `CardTypes` list...
Is there a way to enumerate all possible type( combination)s _a priori_? Then I could add all those to the `CardTypes` type and we wouldn't need to check for partials anymore, making things more realiable.
If enumerating them at coding time is not really possible, I'll have to go back to `card.type` being an unconstrained string with potentially anything in it. Not really pretty.

lusbenjamin: > 205.2. Card Types
>- 205.2a The card types are artifact, conspiracy, creature, enchantment, instant, land, phenomenon, plane, planeswalker, scheme, sorcery, tribal, and vanguard. See section 3, “Card Types.”
>- 205.2b Some objects have more than one card type (for example, an artifact creature). Such objects satisfy the criteria for any effect that applies to any of their card types.

Although WotC never officially says so, they seem to limit themselves to at most 3-4 simultaneous card types on one card. We could use a hard-coded list of (13 choose 3) cardTypeCombos, but I think the partial-matching approach is more intuitive and extensible.

lusbenjamin: @Thaoden pointed out that we may want to more accurately distinguish the super types:
>205.4. Supertypes
>- 205.4a A card can also have one or more supertypes. These are printed directly before its card types. The supertypes are basic, legendary, ongoing, snow, and world.
>- 205.4b An object’s supertype is independent of its card type and subtype, even though some supertypes are closely identified with specific card types.

lusbenjamin: This discussion does highlight that our code here (and possibly in other places) assumes that card types are mutually exclusive, i.e. we only count an "Artifact Creature" as a "Creature" (it occurs first in the `switch`). It would possibly be more correct (and useful?) to count such cards more accurately? (probably best tackled in another focused PR)

Manuel-777: My initial iteration of this, using a set of `if t(ype is in str)` sort of logic was more accurate in that regard, since the first-tested items were sorted on top, thus artifact creature was treated as creature in sort and as both in other checks.

PR #728, opened 2019-11-17
Migrate card types to ts

Thaoden:
Comparing migrations


Thaoden:
Comparing wrong branches :/

PR #729, opened 2019-11-17
Compare Cards List migrations

Thaoden:
Just to compare migrations


src/overlay/overlayUtil.ts
```diff
@@ -5,10 +5,11 @@ import interact from "interactjs";
 import Deck from "../shared/deck";
 
 export interface CardData {
```
Manuel-777: Why is this in overlayUtil? This is a shared type. In my PR I added it to `shared/types/Deck.ts`, but maybe we can add it to a different file that makes more sense.

src/shared/cardsList.ts
```diff
@@ -9,81 +11,83 @@ class CardsList {
    * If an array of IDs is given it sets each quantity to the number of adjacent
    * repetitions
    **/
-  constructor(list = []) {
-    if (!list || list.length === 0) {
-      this._list = [];
-    } else if (typeof list[0] === "object") {
-      this._list = list.map(obj => {
+  private list: CardData[];
```
Manuel-777: I make an "alias" for this as a new type (`v2cardsList` and `v3cardsList`), I feel that way is more organized?

src/shared/cardsList.ts
```diff
@@ -201,16 +214,16 @@ class CardsList {
   getLandsAmounts() {
     var colors = { total: 0, w: 0, u: 0, b: 0, r: 0, g: 0, c: 0 };
```
Manuel-777: I made a type for this return, since it matches the return of `getColorsAmmounts()`

Manuel-777:
I also modified `decks.js` > `decks.ts` that is not included in this PR

src/overlay/overlayUtil.ts
Thaoden: That's a good question. Maybe I just defined it as close as possible to its usage, but I think you're right it should be moved.

lusbenjamin: I put a bunch of types in `overlayUtil` when I stopped being a coward and converted the React Overlay completely into TS. Since most of the main process and shared code was still being converted at the time, I just stuck all of the common metadata interfaces in a common "util" bin. Since most of these _should_ be the same types used in shared and window_main, most of the interfaces int `overlayUtil` should probably move/be merged

PR #730, opened 2019-11-17
Full displays overlay

Manuel-777:
This makes the overlay cover all screens instead of just one like we have it up until now.

Tested on my setup with two displays without any issues (two displays, no GPU)

I wonder if we should default to this method and enable the selector for when it is needed, but then again in a setup without GPU it works fine..

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -437,12 +437,15 @@ function updateOverlayVisibility() {
     clearTimeout(overlayHideTimeout);
     overlayHideTimeout = undefined;
 
-    const { bounds } =
-      electron.screen
-        .getAllDisplays()
-        .find(d => d.id == settings.overlay_display) ||
-      electron.screen.getPrimaryDisplay();
-    overlay.setBounds(bounds);
+    const newBounds = { x: 0, y: 0, width: 0, height: 0 };
+    electron.screen.getAllDisplays().forEach(display => {
+      newBounds.x = Math.min(newBounds.x, display.bounds.x);
+      newBounds.y = Math.min(newBounds.y, display.bounds.y);
+      newBounds.width += display.bounds.width;
+      newBounds.height = Math.max(newBounds.height, display.bounds.height);
+    });
+
+    overlay.setBounds(newBounds);
     overlay.show();
   }
 }
--- a/src/overlay/overlayUtil.tsx
+++ b/src/overlay/overlayUtil.tsx
@@ -116,29 +116,10 @@ export function useEditModeOnRef(
   containerRef: React.MutableRefObject<any>,
   uiScaleFactor: number
 ): void {
-  const bounds = remote.getCurrentWindow().getBounds();
-  const scaledBounds = {
-    x: (bounds.x * 100) / uiScaleFactor,
-    y: (bounds.y * 100) / uiScaleFactor,
-    width: (bounds.width * 100) / uiScaleFactor,
-    height: (bounds.height * 100) / uiScaleFactor
-  };
-  const outerBounds = {
-    left: scaledBounds.x,
-    right: scaledBounds.x + scaledBounds.width,
-    bottom: scaledBounds.y + scaledBounds.height,
-    top: scaledBounds.y
-  };
   const restrictDragBounds: any =
     interact.modifiers &&
     interact.modifiers.restrict({
-      elementRect: { left: 0, right: 1, top: 0, bottom: 1 } as any,
-      restriction: scaledBounds as any
-    });
-  const restrictMaxEdges: any =
-    interact.modifiers &&
-    interact.modifiers.restrictEdges({
-      outer: outerBounds as any
+      elementRect: { left: 0, right: 1, top: 0, bottom: 1 } as any
     });
   useEffect(() => {
     const container = containerRef.current;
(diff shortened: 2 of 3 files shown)


src/main.js
```diff
@@ -437,12 +437,15 @@ function updateOverlayVisibility() {
...
     overlayHideTimeout = undefined;
 
-    const { bounds } =
-      electron.screen
-        .getAllDisplays()
-        .find(d => d.id == settings.overlay_display) ||
-      electron.screen.getPrimaryDisplay();
-    overlay.setBounds(bounds);
+    const newBounds = { x: 0, y: 0, width: 0, height: 0 };
+    electron.screen.getAllDisplays().forEach(display => {
```
Manuel-777: Is there any way the display setup could be vertical? (in which we would need to sum all heights and use the biggest width)

lusbenjamin: I think vertical monitor setups will have their coordinates transformed "upstream" of us by the OS and we can trust that `x` and `y` are the axes we expect. (we should test it anyway if possible)

src/main.js
```diff
@@ -437,12 +437,15 @@ function updateOverlayVisibility() {
...
+      newBounds.x = Math.min(newBounds.x, display.bounds.x);
+      newBounds.y = Math.min(newBounds.y, display.bounds.y);
+      newBounds.width += display.bounds.width;
+      newBounds.height = Math.max(newBounds.height, display.bounds.height);
```
lusbenjamin: why do these behave differently?

Manuel-777: > why do these behave differently?
Since we are calculating the bounding box the the monitors, width needs to be the sum of all widths and height is just the maximum height (this is why I was asking about vertical setups initially) The heights dont need to be added in a horizontal setup.

lusbenjamin: hmmmm. maybe assume our users have a 3x3 grid of monitors, because of course they do?
does this get us closer?
```js
      newBounds.x = Math.min(newBounds.x, display.bounds.x);
      newBounds.y = Math.min(newBounds.y, display.bounds.y);
      newBounds.width = Math.max(newBounds.width, display.bounds.x + display.bounds.width);
      newBounds.height = Math.max(newBounds.height, display.bounds.y + display.bounds.height);
```

src/overlay/overlayUtil.tsx
```diff
@@ -116,29 +116,10 @@ export function useEditModeOnRef(
   containerRef: React.MutableRefObject<any>,
   uiScaleFactor: number
 ): void {
-  const bounds = remote.getCurrentWindow().getBounds();
-  const scaledBounds = {
-    x: (bounds.x * 100) / uiScaleFactor,
-    y: (bounds.y * 100) / uiScaleFactor,
-    width: (bounds.width * 100) / uiScaleFactor,
-    height: (bounds.height * 100) / uiScaleFactor
-  };
```
lusbenjamin: I'm worried that losing this scaling logic will cause a regression

src/main.js
Manuel-777: No, because the `x` from the second monitor is not 0, it will be the width of the first one. and the `x` from the second one the width from the two previous ones, etc. Doing that will give you an exageratedly large display area.
Take for instance this setup (mine actually):
`[
  { bounds: { height: 720, width: 1280, x: 1280, y: 0 } },
  { bounds: { height: 1024, width: 1280, x: 0, y: 0 } }
]`

src/overlay/overlayUtil.tsx
Manuel-777: This was probably lazy by me. But in order for this to work again (restrict the windows within the screen area) we have to synchronize the bounds to the overlay (probably using the `remote` handler and re-runnig hte same calculation as in main)
Do we really really want the overlays restricted?

src/main.js
lusbenjamin: think of it more like we are iterating through all of the screens and checking the absolute coordinates of the top left and bottom right corners. Then we just keep the mins and maxes of those corners.

src/overlay/overlayUtil.tsx
lusbenjamin: we don't want the overlays restricted, but the scaling logic here had to do with correctly handling the "scale factor" setting from the settings page. Won't the screen dimensions will be off by that factor unless we apply this math?

lusbenjamin: nope, I misunderstood. This code is pretty narrow and we don't need it if we are okay giving up the restrictions

src/main.js
lusbenjamin: so for your example case, I would expect the output to be:
` { bounds: { x: 0, y: 0, height: 1024, width: 2560 } }`

Manuel-777: Oh I misunderstood your code (somehow thought it was adding). Yes, the output would be the expected one!

lusbenjamin:
Some other random thoughts:
 - rendering an unusually-large, transparent window across several monitors may have a performance impact
 - as a future optimization we could only "maximize" the transparent space during Edit mode, and then when leaving edit mode attempt to reduce our overlay window size to the minimum bounding box of the overlay windowlets

PR #732, opened 2019-11-18
Overlay borders

Manuel-777:
Just a little detail I want some feedback on; It does make the overlays look more "solid" but on transparency it might be annoying?
image


Manuel-777:
Bonus track!
image
Added an option to change the background color of the overlays and replace default key art :)

lusbenjamin:
😻 love the custom background feature!
👍 for the border, but I can see it not appealing to everyone

Random ideas:
  - have the border color fade along with the transparency setting?
  - do some sort of fancy gradient thingy to only have border weight in the middle of the sides?
  - do some sort of fancy gradient thingy to only have border weight at the corners?
  - only show the border on hover?

Manuel-777:
> * have the border color fade along with the transparency setting?
> * do some sort of fancy gradient thingy to only have border weight in the middle of the sides?
> * do some sort of fancy gradient thingy to only have border weight at the corners?
> * only show the border on hover?

I was thinking to make it fade with the transparency but with a multiplier/inc value so it stands a little like in the screenshots?

I havent thought of fancy gradients, will try!

PR #735, opened 2019-11-18
fix the wildcards fix

Manuel-777:
Did I mention rarity was uppercased?


lusbenjamin:
clearly we are all terrible at being serious 🤣

Thaoden:
Eh, my fix was working! (I thought so, at least...)

PR #756, opened 2019-11-24
better bumped logic

lusbenjamin:
### Motivation
We have some issues with our autogenerated checksum failing, so I want to make the distribution logic as squeaky-clean as we can, just in case.


lusbenjamin:
closing because most of this is now redundant with 87d2f3634309d75011b1a11b1d51d140e1df677b in #786

Issue #758, opened 2019-11-24
Invalid date brakes history tab


riQQ:
The history is completely empty
Version: 2.14.0

```
C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\node_modules\react-dom\cjs\react-dom.development.js:475 Uncaught RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at localTimeSince (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\renderer-util.js:838:43)
    at createDraftTimeDiv (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\renderer-util.js:922:54)
    at attachDraftData (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\renderer-util.js:938:24)
    at DataScroller.renderData (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\history.js:229:39)
    at DataScroller.renderRows (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\data-scroller.js:57:27)
    at DataScroller.render (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\data-scroller.js:27:10)
    at openHistoryTab (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\history.js:153:16)
    at openTab (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\tabControl.js:52:35)
    at clickNav (C:\Users\[user]\AppData\Local\Programs\MTG-Arena-Tool\resources\app.asar\lib\window_main\tabControl.js:138:3)
```

peardox:
Here's mine

[console.log](https://github.com/Manuel-777/MTG-Arena-Tool/files/3887713/console.log)

riQQ:
Fixed in 2.14.1

PR #761, opened 2019-11-25
history page handles bad drafts gracefully

lusbenjamin:


riQQ:
This probably fixes #758, right?

PR #762, opened 2019-11-25
bugfix overlay mis-tracking cards used

lusbenjamin:
### Motivation
- The new log formats now include a sideboard zone in many of the update messages. This caused our "cards used" logic to be wrong for matches with decks that had sideboards.
- Our `CardsList` constructor had a hacky bit of logic to support my old nemesis, the "Group lands hack". It "helpfully" upgraded all zeros to 1, hence messing up the counts for decks with singletons.

https://discordapp.com/channels/463844727654187020/467737642306371584/648570671009628172
image

### DIFF
--- a/src/shared/cards-list.js
+++ b/src/shared/cards-list.js
@@ -15,9 +15,9 @@ class CardsList {
     } else if (typeof list[0] === "object") {
       this._list = list.map(obj => {
         return {
+          quantity: 1, // TODO remove group lands hack
+          id: obj, // TODO remove group lands hack
           ...obj,
-          quantity: obj.quantity || 1,
-          id: obj.id || obj,
           measurable: true
         };
       });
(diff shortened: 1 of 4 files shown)


lusbenjamin:
Original hack was here:
https://github.com/Manuel-777/MTG-Arena-Tool/pull/693/files#diff-55a1417fcd55f1969dae5daf1b120a7cL371-L373

Manuel-777:
> Original hack was here:
> https://github.com/Manuel-777/MTG-Arena-Tool/blob/e063bec3cbe38f362baea8924953ab6d202bee44/src/window_overlay_v3/overlay.jsx#L371

This hack was intended to pass the "chances" from one Deck to the other. If you create a Deck out of a regular internal deck structure it will not search for the `chance` properties within the cards.

A fix could be to teach the `CardsList` class to find `chance` properties and add them if found.

src/shared/cards-list.js
```diff
@@ -15,9 +15,9 @@ class CardsList {
     } else if (typeof list[0] === "object") {
       this._list = list.map(obj => {
         return {
+          quantity: 1, // TODO remove group lands hack
+          id: obj, // TODO remove group lands hack
           ...obj,
-          quantity: obj.quantity || 1,
-          id: obj.id || obj,
```
lusbenjamin: This allows `obj.quantity` to override the default of 1, hence supporting a zero value.

Manuel-777:
While this does seem to fix the issue there are still some improvements to make in the sideboard area and group lands hack, looks g2g!

Issue #763, opened 2019-11-26
Overlay not in line with display when in game using 2k resolution


ClawhammerLobotomy:
Since the update that removed choosing the monitor for the overlay, whenever I launch Arena, the overlays become off-center to my main display, and treat the "game area" as a combination of my other monitors.
This is what I see when in game and using the "Edit overlay positions" function.
![2019-11-25 20_39_01-MTGA](https://user-images.githubusercontent.com/6299800/69592426-d7b19400-0fc3-11ea-85a4-45940d40d766.png)

I've tried using the band-aid fix for "reset windows" from the taskbar.
If I reset the position while in a match, it will stay until I start a new game.

I have 4 monitors set up.
1 at 1920 x 1080
2 at 1080 x 1920 (portrait)
1 at 2560 x 1440 (Main display)

![2019-11-25 20_29_28-Settings](https://user-images.githubusercontent.com/6299800/69592016-9e2c5900-0fc2-11ea-9806-8684d4eee444.png)

Manuel-777:
Hello! In `v2.14.0` I modified the overlays reset position arlgorithm (in settings > overlays) to reset to x=0 and y=0 of the promary/main display, is this the behaviour you see when you reset them?

ClawhammerLobotomy:
Hi Manuel,

When I use the reset, it does go back to (0,0) but it seems to reset to a different position once in a game.

My in-game screenshot shows half the screen as dimmer, which is part of the overlay area of the tracker.
I can also see the overlay on my other monitors in this sort of pattern.
![image](https://user-images.githubusercontent.com/6299800/69593917-bb642600-0fc8-11ea-9672-d3d6c4ead0f7.png)
1. Overlay 1 (not actually able to see it, but I assume it is off screen like this)
2. Overlay 2 (shows in the middle of my secondary monitor
3. Card hover overlay

lusbenjamin:
This is a great amount of detail in a bug report, thank you! Is your tool set to have overlay UI Scale = 100%?

ClawhammerLobotomy:
I work in IT support, so I try to be as detailed as possible.

My UI Scale is set to 80%, but I have tried other scaling which didn't seem to change the behavior.
I just tried with 100% and it does the same.

Thanks

Manuel-777:
It appears from the half-dimmed screenshot we are missing on the calculations to get the full display area and positions.

[...]

PR #768, opened 2019-11-27
update bounds calculation

Manuel-777:
An attempt to fix the bounds algorithm as show fauly in #763 

As show in debug, the setup was:
```
w 1920 h 1080 x 347 y -1080 
w 1080 h 1920 x -1080 y -158
w 1080 h 1920 x  2560 y -158
w 2560 h 1440 x 0 y 0
```
Using our previous code the result was:
```
-1080 -1080 3640 1762
```

While the expected result should be:
```
-1080 -1080 4720 2842
```
Which is what the current output (this PR) is.

In detail, what I do now is calculate the lowest X and Y coodinates fin the first loop and then calculate the longest width and height values using the lowest X and Y as a base for the size calculation.

Also added debug code just in case.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -446,15 +446,25 @@ function updateOverlayVisibility() {
     electron.screen.getAllDisplays().forEach(display => {
       newBounds.x = Math.min(newBounds.x, display.bounds.x);
       newBounds.y = Math.min(newBounds.y, display.bounds.y);
+    });
+    electron.screen.getAllDisplays().forEach(display => {
       newBounds.width = Math.max(
         newBounds.width,
-        display.bounds.x + display.bounds.width
+        Math.abs(newBounds.x) + display.bounds.x + display.bounds.width
       );
       newBounds.height = Math.max(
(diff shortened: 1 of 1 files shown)


lusbenjamin:
Focusing on this test case with the expected result was super useful, as was the debug code. I submitted another suggestion for an algorithm, quickly demonstrated here:
image

I believe the double-loop approach you proposed was correct, but I am not 100% sure that the `Math.abs` call did what you expected it to in all cases (imagine a row of 7 horizontal monitors, 3 will be guaranteed to be "all negative" coordinates).

Manuel-777:
I'm too bad at math to check if they're equivalent, but I'm sure the `abs` method will work in that scenario, since for example the lowest x "-4000" would be added to the furthest right edge (x+w) in absolute

lusbenjamin:
Thought about it some more overnight, and I'm now convinced both new approaches are equivalent. I'm going to revert my commit since there is no discernible performance advantage.

PR #780, opened 2019-12-03
Update proto to MtGA 1.02.01.00

riQQ:
Sadly the diff is so big, because the whitespace was inconsistent before. But appending ?w=1 to a github url makes the differ ignore whitespace so that it is better readable.


Thaoden:
Re: Whitespace diff: You can also click on the gear icon and select "Hide whitespace changes". It resets next time you inspect a diff.

PR #781, opened 2019-12-05
hotfix ensure log IDs are unique

lusbenjamin:
### Motivation
https://discordapp.com/channels/463844727654187020/506793351786266624/651806532651974660
image

### Approach
- add some plumbing to pass around the absolute position of an entry match
- hash the absolute position along with the data to reduce the likelihood of ID collisions (they are still possible with this approach, especially as players build up lots of play history)
- update all places in `labels` where we generate IDs to either use `entry.hash` or a combination of `entry.hash` and the relevent sub-entry data

### DIFF
--- a/src/window_background/arena-log-decoder/arena-log-decoder.js
+++ b/src/window_background/arena-log-decoder/arena-log-decoder.js
@@ -37,10 +37,12 @@ export default function ArenaLogDecoder() {
     let bufferUsed = false;
     let match;
     while ((match = logEntryPattern.exec(buffer))) {
+      const position = bufferDiscarded + match.index;
       const [type, length, entry] = parseLogEntry(
         buffer,
         match[0],
-        match.index
+        match.index,
+        position
       );
       switch (type) {
         case "invalid":
@@ -53,7 +55,7 @@ export default function ArenaLogDecoder() {
           bufferUsed = match.index + length;
           callback({
             ...entry,
-            position: bufferDiscarded + match.index
+            position
           });
           break;
       }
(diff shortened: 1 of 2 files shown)


src/window_background/arena-log-decoder/arena-log-decoder.js
```diff
@@ -37,10 +37,12 @@ export default function ArenaLogDecoder() {
     let bufferUsed = false;
     let match;
     while ((match = logEntryPattern.exec(buffer))) {
+      const position = bufferDiscarded + match.index;
```
Thaoden: The caller in `arena-log-watcher.js` already has a `position` variable. maybe just pass that one in instead of recalculating it.

lusbenjamin: We have to calculate the final absolute position of each JSON entry. The `position` variable in `arena-log-watcher` appears to mean something like: "this is the start position of the last chunk of data I sent to the decoder". Since each chunk could contain any number of valid JSON entries, the final computation of each JSON entry position must happen here inside the `decoder` instead.

My first attempt to derive the JSON positions did indeed rely on plumbing the `watcher`'s `position` value into the `append` function, but that approach failed to account for JSON matches that could possible have started in the previous chunk. This `bufferDiscarded` value seems to get at those edge cases properly, and as an added bonus did not require any of the additional disruptive plumbing. (TBH, I am reading most of this code fore the first time and have almost no context for this multi-tiered buffering system and its former memory leaks.)

Thaoden: fair enough!

PR #784, opened 2019-12-10
Decks react table

lusbenjamin:
### Motivation
Overhauls the Decks page to rebuild the main table widget in React with upgraded controls.
 - SEARCH BAR 🌈 ✨ 😻 ✨ 🦄 
 - Ability to sort and/or filter on any deck attribute or metric we wish to build
 - Ability to show/hide columns
 - More intuitive UI conventions for column sorting/filtering controls
 - Easy to extend with future pagination or async data loading approaches
 - Establish a reasonable approach to convert our current tabular data to React display + control

### Early Demo
decks-react-table-demo

#### New Toys
- [styled-components](https://www.styled-components.com/docs): essentially CSS Modules evolved for React-land. Skips over the need to wait for webpack!
  - https://medium.com/@perezpriego7/css-evolution-from-css-sass-bem-css-modules-to-styled-components-d4c1da3a659b
  - https://medium.com/building-crowdriff/styled-components-to-use-or-not-to-use-a6bb4a7ffc21
- [react-table](https://github.com/tannerlinsley/react-table): does all the heavy lifting on the "control" side of the new decks table. (UI is up to us!)
- [match-sorter](https://github.com/kentcdodds/match-sorter): simple, deterministic, fuzzy-search logic

#### Shared / Non-Decks Code
- Wrapped `relative-time` custom HTML tag in component `RelativeTime` (imitated `LocalTime.tsx`)
- fixed some miscellaneous TS metadata related to Decks

#### TODO
- [x] make controls less ugly
- [x] ~~consider removing (GASP!) the Stats Panel on the right~~ Nah, let's leave it alone.
- [x] remove deprecated CSS styles from `index.css` and `shared.css`
- [ ] ask @Manuel-777 for some spiffy new icons?
- [x] attempt to persist table control settings across page loads using `setLocalState` (necessary hack until entire page is under React control)
- [ ] performance test on slow machine


lusbenjamin:
#### New Duration Metrics
image

#### New Timestamp Metrics
(not shown in screenshot) there are also 2 new optional columns to see last played or last edited independently

src/shared/types/Deck.ts
```diff
@@ -30,6 +30,7 @@ export interface SerializedDeck {
   commandZoneGRPIds?: number[];
   colors?: number[];
   archetype?: string;
+  archived?: boolean;
```
Thaoden: What's the rationale behind this being nullable?

src/window_main/FilterPanel.tsx
```diff
@@ -41,7 +41,7 @@ class FilterPanel {
     showManaFilter: boolean,
     archs: string[],
     showOppManaFilter: boolean,
-    archCounts: { [key: string]: number },
+    archCounts: { [key: string]: number } | null,
```
Thaoden: nit: maybe mark this as `archCounts?: {[key: string]: number}`

src/shared/types/Deck.ts
lusbenjamin: I think an optional field best represents the way we use `archived` in the legacy code. Deck serialization data is only partially complete for parts of its lifespan, which is why all of the other fields that seem "essential" to deck definition are also optional.

lusbenjamin:
#### Uber-Search Bar Demo
decks-react-table-bling

src/window_main/FilterPanel.tsx
lusbenjamin: Since this is in the function signature of the constructor, we cannot use the optional parameter notation unless we move the `archCounts` parameter to the end of the parameters. Although potentially cleaner, it would be a much more disruptive refactor.

package.json
```diff
@@ -55,8 +55,8 @@
     "clean:modules": "rm -fr node_modules",
     "clean:lock": "rm -f package-lock.json",
     "postclean": "npm install",
+    "postinstall": "rm -rf node_modules/@types/react-native",
```
lusbenjamin: Removing `react-native` is necessary to silence `tsc` errors that would otherwise break our build. We do not depend on it, and the redundant types seem to be a result of 3rd party dependencies clashing beyond our control.

src/window_main/aggregator.js
```diff
@@ -150,7 +150,7 @@ class Aggregator {
     return isAfter(new Date(_date), dateFilter);
   }
 
-  _filterDeckByColors(deck, _colors) {
+  static filterDeckByColors(deck, _colors) {
```
lusbenjamin: This should probably move to `renderer-util` or some such common location. (Beyond the scope of this PR.)

src/window_main/components/decks/DecksTable.tsx
```diff
@@ -0,0 +1,498 @@
...
+        return React.useMemo(() => {
+          tableStateCallback(state);
+          const aggFilter = filters.showArchived;
+          const tableFilter = state.filters.archived === "showArchived";
+          if (aggFilter !== tableFilter) {
+            filterMatchesCallback({ ...filters, showArchived: tableFilter });
+          }
+          return state;
+        }, [state, tableStateCallback, filters, filterMatchesCallback]);
+      },
```
lusbenjamin: This is the React-side of the hack to synchronize our React state back to non-React page loads.

src/window_main/DecksTab.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const { decksTable } = getLocalState();
+  mountReactComponent(
+    <DecksTable
+      data={data}
+      filters={filters}
+      cachedState={decksTable}
+      filterMatchesCallback={openDecksTab}
+      tableStateCallback={(state: any): void =>
+        setLocalState({ decksTable: state })
+      }
```
lusbenjamin: This is the non-React-side of the hack to synchronize our React state back to non-React page loads. The current approach is just to shove the `state` object inside our current local state object.

src/shared/types/Deck.ts
Thaoden: Fair enough. I just always struggle with this kind of "Yes, No, Maybe"...

src/window_main/FilterPanel.tsx
Thaoden: Things to do later (tm) ;)

lusbenjamin:
now conflicts with #786, but should probably be stacked on top of it

src/overlay/OverlayWindowlet.tsx
```diff
@@ -142,19 +142,17 @@ export default function OverlayWindowlet(
...
 
   // This needs its own setting, like a checkbox or something
-  const solidBg:boolean = backgroundColor !== "rgba(0,0,0,0)";
-  if (!solidBg)
-    bgStyle.backgroundImage = backgroundImage;
-  else
-    bgStyle.backgroundColor = backgroundColor;
+  const solidBg: boolean = backgroundColor !== "rgba(0,0,0,0)";
+  if (!solidBg) bgStyle.backgroundImage = backgroundImage;
+  else bgStyle.backgroundColor = backgroundColor;
```
Thaoden: I guess that was prettier at work here - but personally I'd keep the bodies on separate lines when using `if - else` (even surround them with braces *duck* )

src/shared/database.ts
```diff
@@ -156,23 +162,27 @@ class Database {
   }
 
   get cardIds(): number[] {
-    return this.cards ? Object.keys(this.cards).map(k => parseInt(k)) : [] as number[];
+    return this.cards
+      ? Object.keys(this.cards).map(k => parseInt(k))
+      : ([] as number[]);
```
Thaoden: Why the `()` ? It is possible to leave out the cast at the end, TS is smart enough to infer it.
Same goes for the next few methods.

src/window_background/data.ts
```diff
@@ -41,38 +39,38 @@ export function createDraft(id: string, entry: any) {
...
+  percentile?: number;
+  leaderboardPlace?: number;
+  commanderGrpIds: any;
 }
 
 export interface MatchData {
   eventId: string;
   matchId: string;
-  InternalEventName?: string,
+  InternalEventName?: string;
```
Thaoden: I guess this one is already used all over the place, but shouldn't it be `internalEventName?` (lowercase i)

src/window_main/ManaFilter.tsx
```diff
@@ -1,62 +1,75 @@
...
+
+  const onClickColorFilter = React.useCallback(
+    (code: ManaFilterKeys) => (event: React.MouseEvent<HTMLDivElement>) => {
+      if (event.currentTarget.classList.contains("mana_filter_on")) {
+        event.currentTarget.classList.remove("mana_filter_on");
+        colors[code] = true;
+      } else {
+        event.currentTarget.classList.add("mana_filter_on");
+        colors[code] = false;
+      }
```
Thaoden:
```suggestion
colors[code] = event.currentTarget.classList.contains("mana_filter_on");
event.currentTarget.classList.toggle("mana_filter_on");
```

src/window_main/components/economy/EconomyDayHeader.tsx
```diff
@@ -13,16 +10,23 @@ import LocalTime from "../../../shared/time-components/LocalTime";
...
 
 function getDayString(daysago: number, timestamp: Date) {
-  return daysago == 0 ? "Today" : daysago == 1 ? "Yesterday" : daysago > 0 ? localDayDateFormat(startOfDay(timestamp)) : '';
+  return daysago == 0
+    ? "Today"
+    : daysago == 1
+    ? "Yesterday"
+    : daysago > 0
+    ? localDayDateFormat(startOfDay(timestamp))
+    : "";
```
Thaoden: I'm a big fan of ternaries, but can we either indent this or decompose it into chained `if`s?

Manuel-777: Thats not something a big fan of ternaries would say..

Thaoden: You don't know this big fan of ternaries ;)

src/window_main/ManaFilter.tsx
lusbenjamin: #791

src/window_main/components/economy/EconomyDayHeader.tsx
lusbenjamin: #791

src/overlay/OverlayWindowlet.tsx
lusbenjamin: #791

src/window_background/data.ts
lusbenjamin: It comes from the naming convention of the serialized data in the Arena logs. At some point, they just starting using capitalization.

src/shared/database.ts
lusbenjamin: I guess it must be a clarity thing?

src/window_background/data.ts
Thaoden: Fair enough.

src/shared/database.ts
Thaoden: I'd remove at least the `()`, maybe even the cast, but that might be personal taste. I don't feel either the parantheses or the cast improve readability.

src/window_main/DecksTab.tsx
```diff
@@ -0,0 +1,188 @@
...
+      ...selected
+    };
+  } else {
+    // default case
+    filters = { ...filters, date: pd.settings.last_date_filter, ...selected };
+  }
+}
+
+//
+export function openDecksTab(_filters = {}): void {
```
Thaoden: nit: rename to `filters` to adhere to standards.

src/window_main/DecksTab.tsx
```diff
@@ -0,0 +1,188 @@
...
+  hideLoadingBars();
+  const mainDiv = resetMainContainer() as HTMLElement;
+  mainDiv.classList.add("flex_item");
+  setFilters(_filters);
+
+  const wrapR = createDiv(["wrapper_column", "sidebar_column_l"]);
+  wrapR.style.width = pd.settings.right_panel_width + "px";
+  wrapR.style.flex = `0 0 ${pd.settings.right_panel_width}px`;
```
Thaoden: nit: reuse the calculated width from the line above

src/window_main/DecksTab.tsx
```diff
@@ -0,0 +1,188 @@
...
+        aggregator.deckStats[id] || getDefaultStats();
+      const avgDuration = Math.round(deckStats.duration / deckStats.total);
+      const recentStats: DeckStats =
+        aggregator.deckRecentStats[id] || getDefaultStats();
+      const winrate100 = Math.round(deckStats.winrate * 100);
+      // compute missing card metrics
+      const missingWildcards = getDeckMissing(deck);
+      const boosterCost = getBoosterCountEstimate(missingWildcards);
+      // compute last touch metrics
+      const lastUpdated = new Date(deck.lastUpdated || NaN);
```
Thaoden: `new Date(NaN)` returns "Invalid date" - is that expected behaviour?

src/window_main/components/decks/DecksTable.tsx
```diff
@@ -0,0 +1,585 @@
...
+  filters: any;
+  filterMatchesCallback: (filters: any) => void;
+  openDeckCallback: (id: string) => void;
+  archiveDeckCallback: (id: string) => void;
+  tagDeckCallback: (deckid: string, tag: string) => void;
+  editTagCallback: (tag: string, color: string) => void;
+  deleteTagCallback: (deckid: string, tag: string) => void;
+  tableStateCallback: (state: any) => void;
+  cachedState: any;
+}
```
Thaoden: nit: maybe push these interfaces in their own file(s)

src/window_main/components/decks/DecksTableComponents.tsx
```diff
@@ -0,0 +1,736 @@
...
+  cell: any;
+  openDeckCallback: (id: string) => void;
+  archiveDeckCallback: (id: string) => void;
+  tagDeckCallback: (deckid: string, tag: string) => void;
+  editTagCallback: (tag: string, color: string) => void;
+  deleteTagCallback: (deckid: string, tag: string) => void;
+}
+
+export const StyledInputContainer = styled.div.attrs(props => ({
+  className: props.className ? props.className : "" + " input_container"
```
Thaoden:
```suggestion
className: props.className || " input_container"
```

Thaoden: If `" input container"` _has_  to be appended, add some parantheses around the ternary:
```suggestion
className: (props.className ? props.className : "") + " input_container"
```

src/window_main/components/decks/DecksTableComponents.tsx
```diff
@@ -0,0 +1,736 @@
...
+  padding-bottom: 4px;
+  &.input_container input {
+    margin: inherit;
+  }
+`;
+
+export const StyledCheckboxContainer = styled.label.attrs(props => ({
+  className: props.className
+    ? props.className
+    : "" + " check_container hover_label"
```
Thaoden: Same as above

src/window_main/components/decks/DecksTableComponents.tsx
```diff
@@ -0,0 +1,736 @@
...
+}
+
+export function NumberRangeColumnFilter({
+  column: { filterValue = [], preFilteredRows, setFilter, id }
+}: {
+  column: any;
+}): JSX.Element {
+  const [min, max] = React.useMemo(() => {
+    let min = preFilteredRows.length ? preFilteredRows[0].values[id] : 0;
+    let max = preFilteredRows.length ? preFilteredRows[0].values[id] : 0;
```
Thaoden: `min === max` here, so instead of re-evaluating, you can just set the variable with `let max = min;`

src/window_main/components/decks/DecksTableComponents.tsx
```diff
@@ -0,0 +1,736 @@
...
+      />
+    </StyledInputContainer>
+  );
+}
+
+export function fuzzyTextFilterFn(
+  rows: any[],
+  id: string,
+  filterValue: string
+): any[] {
```
Thaoden: nit: Maybe type this.

Thaoden: Nit: maybe type this.

src/window_main/components/decks/DecksTableComponents.tsx
```diff
@@ -0,0 +1,736 @@
...
+  return matchSorter(rows, filterValue, { keys: ["values." + id] });
+}
+
+export function uberSearchFilterFn(
+  rows: any[],
+  id: string,
+  filterValue: string
+): any[] {
```
Thaoden: nit: maybe type this, too.

src/window_main/components/decks/DecksTableComponents.tsx
```diff
@@ -0,0 +1,736 @@
...
+    matchSorter(rows, token, {
+      keys: [
+        "values.deckId",
+        "values.name",
+        "values.format",
+        "values.tags",
+        (row: any): string => {
+          const { colors } = row.values;
+          return colors
+            .map((color: number): string => (MANA as any)[color])
```
Thaoden: Why the cast to `any`?

src/window_main/components/decks/DecksTableComponents.tsx
```diff
@@ -0,0 +1,736 @@
...
+  }
+  &.deckTileHover-exit-done {
+    opacity: 0.66;
+    width: 128px;
+  }
+`;
+
+export function ArtTileHeader(): JSX.Element {
+  return <StyledArtTileHeader />;
+}
```
Thaoden: nit: Not sure if this adds value or if it could just be replaced by its return value in callers.

src/window_main/components/decks/DecksTableComponents.tsx
```diff
@@ -0,0 +1,736 @@
+import _ from "lodash";
```
Thaoden: This file is quite long. Is there a sensible way to split it?

lusbenjamin: split into `cells.tsx` and `filters.tsx`.

src/window_main/DecksTab.tsx
lusbenjamin: yes, this is only here to silence TS compile errors since `new Date(undefined)` also returns "Invalid date". Since `lastUpdated` is an optional property, we do the `isValid` check in the lines below to handle invalid dates.

src/window_main/components/decks/DecksTableComponents.tsx
lusbenjamin: to silence TS compile time errors

src/window_main/components/decks/DecksTableComponents.tsx
lusbenjamin: It originally did more interesting things in the render before I added the uber search box. Not sure that this does much anymore other than "implicit memoization" and silencing TS errors related to passing cell props.

src/window_main/components/decks/DecksTableComponents.tsx
Thaoden: That would imply that `MANA` ought to have a different type?

Thaoden:
This looks good to go, as far as I can tell without exactly knowing what is happening inside the `Decks...` classes.
One thing maybe is that `MANA as any` cast, implying that `MANA` could or should be typed differently. But maybe this is something for another PR...

src/window_main/components/decks/DecksTableComponents.tsx
lusbenjamin: I don't know enough about TS to be sure. It's also possible that I could do some elaborate manipulation of the `colors` map-like object here to convince TS that it only contains `const` keys? (The "proper" approach is beyond my experience in this situation. I'm not even sure I understand all of the possible approaches.)

PR #791, opened 2019-12-13
review feedback

lusbenjamin:
transplanted from #784


Thaoden:
Should someone else approve these additionally?

PR #793, opened 2019-12-15
Bugfix collection stats performance

lusbenjamin:
### Motivation
The collection tab statistics sub-page is one of the coolest displays in the tool. Sadly, opening the page takes about 5 billion years on a slow machine. This PR does 2 things:
- Extracts all of the code related to the sub-page into a separate module and upgrades it to TS
- Optimizes the `getCollectionStats` function by splitting apart the nested loop related to wanted cards.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -378,7 +378,7 @@ export function getMissingCardCounts(deck) {
   );
   allCardIds.forEach(grpid => {
     const missing = getCardsMissingCount(deck, grpid);
-    if (missing) {
+    if (missing > 0) {
       missingCards[grpid] = missing;
     }
   });
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -371,6 +371,20 @@ export function getCardsMissingCount(deck, grpid) {
   return mainMissing + sideboardMissing;
 }
 
+export function getMissingCardCounts(deck) {
+  const missingCards = {};
+  const allCardIds = new Set(
+    [...deck.mainDeck, ...deck.sideboard].map(card => card.id)
+  );
+  allCardIds.forEach(grpid => {
+    const missing = getCardsMissingCount(deck, grpid);
+    if (missing) {
+      missingCards[grpid] = missing;
+    }
+  });
+  return missingCards;
+}
+
 export function getBoosterCountEstimate(neededWildcards) {
   let boosterCost = 0;
   const boosterEstimates = {
--- a/src/window_main/collection.js
+++ b/src/window_main/collection.js
@@ -1,13 +1,9 @@
-import anime from "animejs";
+/* eslint-disable max-statements, complexity */
 import { remote } from "electron";
 const Menu = remote.Menu;
 const MenuItem = remote.MenuItem;
 
-import {
-  COLORS_BRIEF,
-  CARD_RARITIES,
-  EASING_DEFAULT
-} from "../shared/constants";
+import { COLORS_BRIEF } from "../shared/constants";
 import db from "../shared/database";
 import pd from "../shared/player-data";
 import { queryElements as $$, createDiv } from "../shared/dom-fns";
@@ -16,41 +12,16 @@ import { addCardHover, attachOwnerhipStars } from "../shared/card-hover";
 import {
   collectionSortRarity,
   getCardImage,
-  getCardsMissingCount,
   openScryfallCard,
   replaceAll
 } from "../shared/util";
-import Colors from "../shared/colors";
-import {
-  MULTI,
-  COLORLESS,
-  WHITE,
-  BLUE,
-  BLACK,
-  GREEN,
-  RED
-} from "../shared/constants.js";
-import {
-  hideLoadingBars,
-  changeBackground,
-  ipcSend,
-  resetMainContainer
-} from "./renderer-util";
+import { hideLoadingBars, ipcSend, resetMainContainer } from "./renderer-util";
+import { openSetStats } from "./collectionStats";
 
 let collectionPage = 0;
 let sortingAlgorithm = "Sort by Set";
 let filteredSets;
 let filteredMana;
-let orderedSets;
-
-const ALL_CARDS = "All cards";
-const SINGLETONS = "Singletons (at least one)";
-const FULL_SETS = "Full sets (all 4 copies)";
-const BOOSTER_CARDS = "Only in Boosters";
-
-let countMode = ALL_CARDS;
-let displayMode = BOOSTER_CARDS;
-let defaultSetName = "Complete collection";
 
 //
 function get_collection_export(exportFormat) {
@@ -110,35 +81,6 @@ function collectionSortSet(a, b) {
 }
 
 //
-class CountStats {
-  constructor(
-    owned = 0,
-    total = 0,
-    unique = 0,
-    complete = 0,
-    wanted = 0,
-    uniqueWanted = 0,
-    uniqueOwned = 0
-  ) {
-    this.owned = owned;
-    this.total = total;
-    this.unique = unique;
-    this.complete = complete; // all 4 copies of a card
-    this.wanted = wanted;
-    this.uniqueWanted = uniqueWanted;
-    this.uniqueOwned = uniqueOwned;
-  }
-
-  get percentage() {
-    if (this.total) {
-      return (this.owned / this.total) * 100;
-    } else {
-      return 100;
-    }
-  }
-}
-
-//
 function collectionSortName(a, b) {
   a = db.card(a);
   b = db.card(b);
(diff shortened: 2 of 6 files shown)


lusbenjamin:
#### Demo of working svg icons
This depends on https://github.com/mtgatool/MTG-Arena-Tool-Metadata/pull/1
image

src/shared/util.js
```diff
@@ -371,6 +371,20 @@ export function getCardsMissingCount(deck, grpid) {
...
+export function getMissingCardCounts(deck) {
+  const missingCards = {};
+  const allCardIds = new Set(
+    [...deck.mainDeck, ...deck.sideboard].map(card => card.id)
+  );
+  allCardIds.forEach(grpid => {
+    const missing = getCardsMissingCount(deck, grpid);
+    if (missing) {
+      missingCards[grpid] = missing;
+    }
```
Thaoden:
```suggestion
missingCards[grpid] = getCardsMissingCount(deck, grpid);
```

`getCardsMissingCount` always returns a number, so `if(missing)` always passes.

src/window_main/collectionStats.ts
```diff
@@ -0,0 +1,521 @@
...
+
+const ALL_CARDS = "All cards";
+const SINGLETONS = "Singletons (at least one)";
+const FULL_SETS = "Full sets (all 4 copies)";
+const BOOSTER_CARDS = "Only in Boosters";
+
+let countMode = ALL_CARDS;
+let displayMode = BOOSTER_CARDS;
+let defaultSetName = "Complete collection";
+const playerData = (pd as unknown) as any;
```
Thaoden: Funny cast..?

src/window_main/collectionStats.ts
```diff
@@ -0,0 +1,521 @@
...
+      stats.complete[card.rarity].wanted += wanted;
+      // count unique cards we know we want across decks
+      stats[card.set][card.rarity].uniqueWanted += Math.min(1, wanted);
+      stats.complete[card.rarity].uniqueWanted += Math.min(1, wanted);
+      obj.wanted = wanted;
+    }
+    if (!stats[card.set].cards[colorIndex])
+      stats[card.set].cards[colorIndex] = {};
+    if (!stats[card.set].cards[colorIndex][card.rarity])
+      stats[card.set].cards[colorIndex][card.rarity] = [];
```
Thaoden: You could prevent these properties being null / undefined when using classes with a default constructor that fully initialize to default objects, instead of interfaces.

src/window_main/collectionStats.ts
```diff
@@ -0,0 +1,521 @@
...
+  substats.innerHTML = "";
+
+  const label = document.createElement("label");
+  label.innerHTML = setName + " completion";
+  substats.appendChild(label);
+
+  // Draw completion table for this set
+  if (setName != "Complete collection") {
+    const table = createDiv(["completion_table"]);
+    for (let color = 0; color < 7; color++) {
```
Thaoden: You could add a `ALL_COLOURS` array or dictionary or record to `constants.ts` and `foreach` it.

src/window_main/collectionStats.ts
```diff
@@ -0,0 +1,521 @@
...
+      );
+      compDiv.style.opacity = "1";
+      substats.appendChild(compDiv);
+    }
+    wanted[rarity] = countStats.wanted;
+    missing[rarity] = countStats.total - countStats.owned;
+  });
+
+  // If the set has a collationId, it means boosters for it exists
+  if (db.sets[setName] && db.sets[setName].collation) {
```
Thaoden:
```suggestion
if(db.sets[setName]?.collation){
```

src/window_main/collectionStats.ts
```diff
@@ -0,0 +1,521 @@
...
+      const rareWantedSpan = document.createElement("span");
+      rareWantedSpan.innerHTML = `<i>~${chanceBoosterRareWanted} chance next booster has ${wantedText} rare.</i>`;
+      rareWantedSpan.style.fontSize = "13px";
+      rareWantedIcon.appendChild(rareWantedSpan);
+      rareWantedDiv.appendChild(rareWantedIcon);
+      substats.appendChild(rareWantedDiv);
+    }
+
+    // chance that the next booster opened contains a mythic missing from one of our decks
+    const possibleMythics =
```
Thaoden: This block is almost duplicated from `const possibleRares` - any possibility to extract common things into a function?

src/window_main/collectionStats.ts
```diff
@@ -0,0 +1,521 @@
+import anime from "animejs";
```
Thaoden: This file is quite long. Any possibility to split the rendering part from the "library-only" part?

lusbenjamin:
As usual, @Thaoden has some excellent feedback! 😸 

🤔 From a tactics perspective, I think I'd prefer to keep this PR aimed solely at improving the performance of the page and tackle more wholesale code quality upgrade in a separate refactor. (Ideally, I would have only patched the algorithm in this PR, but I couldn't help myself from splitting the file and moving the part I was touching to TS.)

src/window_main/collectionStats.ts
lusbenjamin: Absolutely! I would recommend doing this split as part of a migration into React components.

src/window_main/collectionStats.ts
lusbenjamin: Some kind of bogus cast happens in any TS file that consumes the `player-data` module. They should all go away when one of us is brave enough to type-convert that central puzzle piece.

src/shared/util.js
lusbenjamin: True, but I actually intended this one in the arithmetic sense. Updating the logic for clarity.

src/window_main/collection.js
```diff
@@ -147,118 +89,13 @@ function collectionSortName(a, b) {
...
-    let col = new Colors();
-    col.addFromCost(card.cost);
-    let colorIndex = col.getBaseColor();
-
-    // count cards we know we want across decks
-    const wanted = Math.max(
-      ...pd.deckList
-        .filter(deck => deck && !deck.archived)
-        .map(deck => getCardsMissingCount(deck, card.id))
-    );
```
lusbenjamin: This is the extremely inefficient part of the old algorithm. I can't believe some hack would think it was okay to repeat this calculation for each card, causing it to loop over (all cards in db X all decks X all cards in each deck). What sort of person thinks that O(n) is remotely acceptable? Let's take a look at git blame and make this into a teaching moment!

... aw crap. 😅

src/window_main/collectionStats.ts
```diff
@@ -0,0 +1,521 @@
...
+function getCollectionStats(): { [key: string]: SetStats } {
+  const wantedCards: { [key: string]: number } = {};
+  pd.deckList
+    .filter(deck => deck && !deck.archived)
+    .forEach(deck => {
+      const missing = getMissingCardCounts(deck);
+      Object.entries(missing).forEach(([grpid, count]) => {
+        wantedCards[grpid] = Math.max(wantedCards[grpid] ?? 0, count);
+      });
+    });
```
lusbenjamin: This is the new computation _outside_ of the main loop. It uses a new `util` function, because we didn't have enough similarly named missing-card-counters yet.

src/window_main/collectionStats.ts
Thaoden: `pd` currently doesn't have a `cards` property. Adding that to the `playerData` type might already be enough that this cast becomes unneccessary.

src/shared/util.js
Thaoden: I see the check for `> 0` in the new version of the file - actually, you don't need that. Either `getMissingCardCount === 0`, in which case it is perfectly fine to set that to `missingCards[grpid]` (it is the default value anyway), or you want to set the value you just computed :)

Thaoden:
You flatter me ;)

The things I suggested are in no way blocking issues and some of them probably would need rather drastic changes (like converting from interface to class).

What I really dislike though is the `pd` cast (though not sure if it wasn't already present before). _If possible_, at least that one should be resolved (but, as I said, nothing is really blocking).

src/window_main/collectionStats.ts
lusbenjamin: I think you may be underestimating the jank that goes on inside `player-data`. I agree that it is one of the highest priority areas to attempt to introduce types, but I also suspect that "typing `player-data` correctly" ~= "typing the entire data model correctly" and that it will be difficult to improve incrementally.

src/shared/util.js
lusbenjamin: The goal here is to produce a sparsely populated object to make looping over missing cards more efficient. We only want to add non-zero missing cards to the result.

src/window_main/collectionStats.ts
lusbenjamin: Solved by adding a property declaration to the constructor!

src/shared/util.js
Thaoden: Didn't see that `missingCards` is a dictionary and not an object, you're right of course!

Thaoden:
Looks good! Some todos left for later work :)

PR #799, opened 2019-12-20
React Decks Improvements

Manuel-777:
I will be dumping whatever improvements to the new decks page here as I figure how it works :)

### Bugfixes / Improvements / TODO
- sort icon
- performance optimizations for art tile
- row hover tile animation _so smooth_
- row-click handler instead of tile/name-click handler
- better tooltips for tags
- upgrade to latest release-candidate of `react-table`
- bugfix custom decks not appearing
- misc code cleanup
- make filter toggle affect column width less


Manuel-777:
Some things I noticed;
- The styles are a mess. Styled components are nice but I think it adds a lot of unnecessary clutter to the code and makes it super hard to follow.. Its also not as cool as webpack's styled components?
Part of why it confuses me is we are already using a mix of classes, styled components and inline styles. Some times one has to track a single div's style transformations in three different places.

- I still have no idea how the table/sorts/filters and overall filter magic works. Im sure the components themselves can be organized, separated and named accordingly to make our life's better, but I have to figure the former before the later.

lusbenjamin:
On styled components vs CSS modules:
>styled-components: essentially CSS Modules evolved for React-land. Skips over the need to wait for webpack!
> - https://medium.com/@perezpriego7/css-evolution-from-css-sass-bem-css-modules-to-styled-components-d4c1da3a659b
> - https://medium.com/building-crowdriff/styled-components-to-use-or-not-to-use-a6bb4a7ffc21

Issue #653, opened 2019-10-22
Overlays should always show their current status


AnnanFay:
Overlays should never be empty, regardless of why they are empty (lack of data / user not logged in / etc). They should also not contain stale information.

For instance, if the draft overlay is set to be persistently on but there is not a draft in progress then there should be a message letting the user know that an empty overlay is completely normal. For every case where the overlay is currently empty we can instead show a message explaining why there is no data.

Likewise, after drafting is done the old data should be removed from the overlay so it doesn't contain information from a previous draft.

Manuel-777:
I really like this idea! we shoud list all possible states (if _this_ or _that_ then _what_) that are confusing/not clear and their respective messages. I made a little CSS to test;
![image](https://user-images.githubusercontent.com/4711391/67339677-0f7b7680-f502-11e9-8366-81aa179c3ea0.png)

PR #804, opened 2019-12-21
pre-populate draft overlays with last draft

lusbenjamin:
This PR adds logic to pre-populate the draft-mode overlays with data from the last complete draft. This makes the "Always show overlay" option much more useful when attempting to tune the HUD outside of a draft. (It is also a superb debugging and development tool to avoid spending 5k gold.)

### DIFF
--- a/src/window_background/loadPlayerConfig.js
+++ b/src/window_background/loadPlayerConfig.js
@@ -1,7 +1,7 @@
 import _ from "lodash";
 import { shell } from "electron";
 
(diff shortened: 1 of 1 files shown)


lusbenjamin:
This does not really solve #653 but is somewhat related.

Manuel-777:
> This does not really solve #653 but is somewhat related.

Its actually the oposite to what the issue says? Instead of showing old data there should be a status message as of why there is no new draft data.

It should help us with viewwing problems in the draft overlays trough, so ill keep it until we do that.

PR #805, opened 2019-12-21
Port background-util.js to TS

Thaoden:
This is just a small port from `background-util.js` to TypeScript.


src/window_background/backgroundUtil.ts
```diff
@@ -105,23 +105,28 @@ export function getDateFormat(dateStr) {
   }
 }
 
-export function normaliseFields(iterator) {
-  if (typeof iterator == "object") {
-    return _.transform(iterator, function(result, value, key) {
-      let nkey =
-        typeof key == "string" ? key.replace(/List$/, "").toLowerCase() : key;
```
lusbenjamin: This logic did not make it into the TS conversion. Was that a deliberate change?

Manuel-777: This function was only useful in the payload decoder, Which was only used to get the sideboard changes on BO3, does that still work with these changes?

Thaoden: TS types told me that `key` is always of type `number`, so the first part of the ternary didn't ever evaluate, thus I removed it.

Manuel-777: pushed c5e5d933b714522a450eb3b15c4d7917ff412f49 with a "fix" for this, it makes TS happy without changing the functionality.

PR #807, opened 2019-12-21
Migrate getOpponentDeck to TS and move the color archetype check into…

Thaoden:
… colors

This migrates `getOpponentDeck` to TypeScript. 

At the same time `getColorArchetype()` is moved into class `colors` (where it is better placed, I think). Also, the archetype check changed from a whole lot of repeated `if`-checks to a flag based implementation.

In `getOpponentDeck.ts` on lines 18-19, the check is
```
    (format !== "Standard" && format !== "Traditional Standard") ||
    globals.currentMatch.oppArchetype == "Unknown"
```

I'm pretty sure the first part will always evaluate to `false` given the logical AND, if someone could confirm, I'll remove that line.


lusbenjamin:
- ❤️ moving `getColorArchetype` into the `colors` module
- I think the bit-wise arithmetic is super-cool, but that refactor probably makes it harder for most programmers to understand.
- Re: lines 18-19 check, `format` can basically be any string (I think no whitespace allowed). This code needs to handle "Historic", "Brawl", "Artisan", etc.

Thaoden:
Should I remove the flags and reimplement the `if`-checks, then?

Manuel-777:
> Should I remove the flags and reimplement the `if`-checks, then?

I think its okay, the whole purpose of the Colors class is to allow arithmetics like that :)

PR #812, opened 2019-12-22
Migrate card hover to ts

Thaoden:
Migrate `card-hover.js` to TypeScript.

I left some additional comments below.


src/shared/cardHover.tsx
```diff
@@ -6,52 +6,63 @@ import { queryElements as $$ } from "./dom-fns";
...
-    $$(".loader, .main_hover").forEach(element => (element.style.opacity = 1));
+    $$(".loader, .main_hover").forEach(
+      (element: any) => (element.style.opacity = "1")
+    );
     // Split cards are readable both halves, no problem
-    if (card.dfc == FACE_DFC_BACK || card.dfc == FACE_DFC_FRONT) {
-      $$(".loader_dfc, .main_hover_dfc").forEach(el => {
-        show(el);
+    if (
+      dbCard.dfcId &&
```
Thaoden: This check lets TS infer that `dbCard.dfcId` cannot be `false` in line 44.

src/shared/cardHover.tsx
```diff
@@ -64,20 +75,14 @@ export function addCardHover(element, card) {
   element.addEventListener("mouseleave", hideHover);
 }
 
-function show(element, mode) {
```
Thaoden: I have inlined `show` and `hide`, both were called at a single point only and were basically one-liners.

src/window_main/components/economy/EconomyRow.tsx
```diff
@@ -434,7 +434,7 @@ interface InventoryCardProps {
 
 function InventoryCard(props: InventoryCardProps) {
   const { card, isAetherized, quantity } = props;
-  const inventoryCardRef = React.useRef(null);
+  const inventoryCardRef = React.useRef<HTMLDivElement>(null);
```
Thaoden: The use of the generic `HtmlDivElement` is necessary here to enable the cast to `HtmlElement` on line 449 without having to cast to `unknown` first and to enable the assignment in line 458 (`ref` expects `HtmlDivElement | null | ...`, so using the generic `HtmlElement` here wouldn't work.

src/shared/cardHover.tsx
Thaoden: Not sure if that makes the second part of the check (line 37) obsolete, though?

PR #814, opened 2019-12-22
Deck tables styling

Manuel-777:
Some CSS love for the deck tables.

Instead of using classic HTML Tables I opted for a CSS approach and removed tons of useless `div` floating around.

### DIFF
--- a/src/window_background/arena-log-decoder/arena-log-decoder.js
+++ b/src/window_background/arena-log-decoder/arena-log-decoder.js
@@ -1,6 +1,6 @@
 import _ from "lodash";
 import nthLastIndexOf from "./nthLastIndexOf";
-import * as jsonText from "./json-text";
+import * as jsonText from "./jsonText";
 import sha1 from "js-sha1";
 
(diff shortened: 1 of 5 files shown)


src/window_background/arena-log-decoder/arena-log-decoder.js
```diff
@@ -126,7 +126,7 @@ function parseLogEntry(text, matchText, position, absPosition) {
     ];
   }
 
-  for (let pattern of LABEL_JSON_PATTERNS) {
+  for (const pattern of LABEL_JSON_PATTERNS) {
```
Thaoden: Does this work? I thought marking this as const makes it immutable, so it wouldn't change its value when looping...

Manuel-777: That was actually EsLint not me, I guess it re-declares the variable on each loop?

Thaoden: Just tried it with a simple "print each value of this 1..10 array" loop, indeed it does...

PR #815, opened 2019-12-23
GRE to Client Interpreter and dependecies

Manuel-777:
BIG migration ahead!

Still some linting issues and probably a lot of code reordering to be done; Some things im not very happy with:
- The types probably need to be organized in more sub-files, to make it easier to find them and import later, which we will be doing a lot once this begins to scale up and spread into other files, like `labels.js`. We have to decide how we want to organize these before taking that step.
- function `getCardsTypeZone` L585 has a *very* difficult form. Its a ternary array of keys (if that makes any sense), I didnt wanted to give up and type it as :any because that felt bad. But maybe we should?
- I think some types can be extensions of others? I noticed this may be true too far into the refactoring , I still tried to refer to the same types instead of creating new ones whenever possible.

When in doubt I CTRL F a key in an output log and copy pasted the whole object into a type. Hopefully we can map everything like that and stay as truthful as possible with what is being read.


src/window_background/getNameBySeat.ts
```diff
@@ -8,7 +8,7 @@ function getPlayerNameWithoutSuffix(playerName: string) {
 }
 
 // Get player name by seat in the game
-const getNameBySeat = function(seat: 1 | 2): string {
+const getNameBySeat = function(seat: number): string {
```
Thaoden: Judging from the calls, `seat` can only be `1` or `2` - I'd leave the type so narrowly constrained, though maybe we can add consts for it (`ON_THE_PLAY = 1`, `ON_THE_DRAW = 2` or sth like it)

src/window_background/gre-to-client-interpreter.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-console */
```
Thaoden: nit: rename to `greToClientInterpreter.ts` to remove kebap case in file names

src/window_background/gre-to-client-interpreter.ts
```diff
@@ -38,32 +51,19 @@ function changePriority(previous, current, time) {
   globals.currentMatch.currentPriority = current;
 }
 
-const actionLogGenerateLink = function(grpId) {
-  var card = db.card(grpId);
-  return '<log-card id="' + grpId + '">' + card.name + "</log-card>";
+const actionLogGenerateLink = function(grpId: number): string {
+  const card = db.card(grpId);
+  if (card) {
```
Thaoden: nit: could be replaced by a ternary (but I guess that's personal taste)

src/window_background/gre-to-client-interpreter.ts
```diff
@@ -110,20 +137,22 @@ function keyValuePair(obj, addTo) {
...
-function processAnnotations() {
-  globals.currentMatch.annotations.forEach(ann => {
+function processAnnotations(): void {
+  globals.currentMatch.annotations.forEach((ann: AnnotationType) => {
     // if this annotation has already been processed, skip
     if (globals.currentMatch.processedAnnotations.includes(ann.id)) return;
 
-    let details = {};
+    let details: DetailsType;
     if (ann.details) {
```
Thaoden: This check is now obsolete due to the type annotation in line 141.

src/window_background/gre-to-client-interpreter.ts
```diff
@@ -110,20 +137,22 @@ function keyValuePair(obj, addTo) {
...
-      ann.details.forEach(detail => (details = keyValuePair(detail, details)));
+      ann.details.forEach(
+        (detail: DetailsType) => (details = keyValuePair(detail, details))
+      );
     }
 
     let processedOk = true;
```
Thaoden: You can pull line 168 up into the `try`-block and get rid of `processedOk`.

src/window_background/gre-to-client-interpreter.ts
```diff
@@ -391,30 +438,36 @@ annotationFunctions.AnnotationType_DamageDealt = function(ann, details) {
...
+annotationFunctions.AnnotationType_ModifiedLife = function(
+  ann: AnnotationType,
+  details: DetailsType
+): void {
+  const affected = ann.affectedIds[0];
+  const total = globals.currentMatch.players[affected].lifeTotal;
 
-  if (details.life > 0) details.life = "+" + details.life;
+  let lifeStr = "";
+  if (details.life > 0) lifeStr = "+" + details.life;
```
Thaoden: nit: these two lines could be combined into a ternary.

src/window_background/gre-to-client-interpreter.ts
```diff
@@ -427,24 +480,30 @@ annotationFunctions.AnnotationType_TargetSpec = function(ann) {
...
+  const top = details.topIds;
+  const bottom = details.bottomIds;
 
-  let top = details.topIds;
-  let bottom = details.bottomIds;
+  let newTop: number[] = [];
+  let newBottom: number[] = [];
   if (!Array.isArray(top)) {
-    top = top !== undefined ? [top] : [];
+    newTop = top !== undefined ? [top] : [];
```
Thaoden: `top` comes from `details.topIds` which is typed as `number`, so it can never be `undefined`. Same goes for `bottom`.
If some JS code doesn't care about the types and sets it to `undefined`, extend the type def accordingly.

src/window_background/gre-to-client-interpreter.ts
```diff
@@ -773,24 +735,30 @@ function checkForStartingLibrary() {
     }
   });
 
+  // Probably just escape valves?
   if (globals.currentMatch.gameStage !== "GameStage_Start") return -1;
```
Thaoden: This doesn't compile - return type is expected as `number[]`

src/window_background/gre-to-client-interpreter.ts
```diff
@@ -857,16 +825,95 @@ function checkTurnDiff(turnInfo) {
...
+  forceDeckUpdate();
+  update_deck(false);
+};
+
+GREMessages.GREMessageType_DieRollResultsResp = function(msg): void {
   if (msg.dieRollResultsResp) {
-    let highest = msg.dieRollResultsResp.playerDieRolls.reduce((a, b) => {
-      if (a.rollValue > b.rollValue) {
+    const highest = msg.dieRollResultsResp.playerDieRolls.reduce((a, b) => {
+      if ((a.rollValue ? a.rollValue : 0) > (b.rollValue ? b.rollValue : 0)) {
```
Thaoden:
```suggestion
if((a.rollValue ?? 0) > (b.rollValue ?? 0)){
```

src/window_background/types/greInterpreter.ts
```diff
@@ -0,0 +1,113 @@
+import anyCardsList from "../../shared/deck";
+import { GameInfo, TurnInfo, PlayerData } from "./currentMatch";
+
+export interface KeyValuePair {
```
Thaoden: What's the use case for this KVP?

src/window_background/getNameBySeat.ts
Manuel-777: I would expect this to be any number if WoTc ever implements games from more than two players

src/window_background/types/greInterpreter.ts
Manuel-777: Its how values are shown in the logs, instead of listing each value case for each different object structure we just read a KVP that tells use what are we reading. Key is the "key" name, like a JSON and type is the value type that is then used to read the value.

 Its probably inherithed from the types Arena uses in C#.

src/window_background/gre-to-client-interpreter.ts
Manuel-777: I did not want to change the functionality here, this codes arent mine and I first wanted to convert the whole file before digging into why this function works like that.

src/window_background/gre-to-client-interpreter.ts
Manuel-777: I have to see a Scry Annotation to be sure of the numbers passed on in the details and make the changes. It seems I coded it this way expecting top to be either a number or an array.

src/window_background/getNameBySeat.ts
Thaoden: If ever they implement that ;)

src/window_background/gre-to-client-interpreter.ts
Thaoden: You could type the return value as `number[] | number` then.

src/window_background/types/greInterpreter.ts
Thaoden: Reason I'm asking is why not use implemented types like `{[key:string]:number[]}`? Ah, because MTGA logs use the property names `valueXXX`?

src/window_background/greToClientInterpreter.ts
```diff
@@ -72,11 +70,40 @@ const gameObjectCardTypes = [
...
+    globals.idChanges[instanceID]
+  ) {
+    instanceID = globals.idChanges[instanceID];
+  }
+
+  const instance = globals.currentMatch.gameObjs[instanceID];
+  if (instance) {
+    return instance;
+  }
+  throw new NoInstanceException(orig, instanceID, instance);
```
Thaoden: `instance` is `null` or `undefined` here - is there a reason to pass it into the exception constructor (though I don't have its signature handy...)?

src/window_background/types/greInterpreter.ts
```diff
@@ -35,27 +35,145 @@ interface ValueType {
...
 
 export interface GameObjectType {
+  type: string;
+  zoneId: number;
+  controllerSeatId: number;
+  grpId: number;
+  instanceId: number;
+  ownerSeatId: number;
+  /*
+  groupId: number;
```
Manuel-777: This is still a WIP until I map the rest of the object types, so I kept some repeated declarations between `GameObjectType` and its children. All keys that are repeated across the children should be here and not in the children for simplicity.

Thaoden: Design question: Should the types be prefixed (`type = GameObjectType_{type}`) or not (`type = {type}`) because the type exists on an object of supertype `GameObjectType` anyway?

Really like the move to Discriminating Unions!

Manuel-777: Not sure, my naming conventions are pretty bad.. for the string discriminators they need to match the logs, but interfaces can be anything

src/window_background/greToClientInterpreter.ts
Manuel-777: Yeah I dont think it will ever be defined..

PR #819, opened 2019-12-23
Migrate whats-new to TypeScript

Thaoden:
Migrate `whats-new.js` to TypeScript.

`prev` and `next` moved closer to where they're actually used, this also prevents TS thinking they might be undefined.


Manuel-777:
huh, this is a cute little deprecated code..

Thaoden:
:D I'm happy to remove it instead of migrating it...

PR #820, opened 2019-12-24
Fix typing of db.card()

Thaoden:
Previously, `db.card()` would return `false` when a value was not available, resulting in a bunch of weird checks necessary before accessing the card object. This changes the signature to return `undefined` instead of `false`.

Additionally, the same pattern was used with `dfcId`, this property was marked optional instead of being potentially `false`.

A bunch of checks have been changed to use the new types.

Also, the input signature of `db.card()` was changed to accept an optional `cardId` and returning `undefined` when none was provided, in order to make patterns like `if (card.dfcId !== undefined) obj = db.card(card.dfcId)` obsolete.


Manuel-777:
Nice ! I came to the same conclusion in the website;
https://github.com/mtgatool/mtgatool-web/blob/6ca15b73c475263bc4a1bb85eaac07d444bfaed9/src/shared/database.js#L36

I really like the use of the `card?,key` syntax here, I wasnt aware of it !

Thaoden:
I think though the tool metadata uses `false` for `dfcId`, so that might need to be updated accordingly...

PR #821, opened 2019-12-28
Background Labels refactor

Manuel-777:
HUGE REFACTOR INCOMING!

I finally did it and typed the log entries we use, also separated the functions used in `labels.js` into individual functions we can reuse.

Still some things left to do, like test drafts and actually making types for them, is probably the biggest one I can remember now. Most things I was not happy with I tagged with a `REVIEW` comment.

One file I could not type properly (make lint happy) was `arena-log-decoder`, I even consireded leaving it as js due to the ammount of functions being passed around.


src/types/log.ts
```diff
@@ -0,0 +1,174 @@
...
+  | GameStart
+  | GameStop
+  | Connected
+  | SceneChange;
+
+export interface LogInfo {
+  jsonrpc: string;
+  method: string;
+  params: LogParams;
+  id: "193";
```
lusbenjamin:
```suggestion
  id: number;
```

src/types/match.ts
```diff
@@ -0,0 +1,68 @@
+import { Result } from "./greInterpreter";
+
+interface ReservedPlayer {
+  userId: string;
+  playerName: string;
+  systemSeatId: 1;
```
lusbenjamin:
```suggestion
  systemSeatId: number;
```

src/types/match.ts
```diff
@@ -0,0 +1,68 @@
...
+  teamId: 1;
```
lusbenjamin:
```suggestion
  teamId: number;
```

src/window_background/onLabel/ClientToMatchServiceMessageTypeClientToGREMessage.ts
```diff
@@ -0,0 +1,84 @@
...
+  type: string;
+}
+
+interface Entry extends LogEntry {
+  json: () => Payload;
+}
+
+function decodePayload(payload: any, msgType: string): any {
+  const messages = require("../messages_pb");
+  const binaryMsg = new Buffer.from(payload, "base64");
```
lusbenjamin:
```suggestion
  const binaryMsg = Buffer.from(payload, "base64");
```

src/window_background/onLabel/ClientToMatchServiceMessageTypeClientToGREMessage.ts
```diff
@@ -0,0 +1,84 @@
...
+  };
+  type: string;
+}
+
+interface Entry extends LogEntry {
+  json: () => Payload;
+}
+
+function decodePayload(payload: any, msgType: string): any {
+  const messages = require("../messages_pb");
```
lusbenjamin: just curious, why do we have to do this here instead of as an `import`?

src/window_background/onLabel/EventMatchCreated.ts
```diff
@@ -0,0 +1,51 @@
...
+  json: () => EntryJson;
+}
+
+export default function EventMatchCreated(entry: Entry): void {
+  const json = entry.json();
+  if (!json) return;
+  const matchBeginTime = globals.logTime || new Date();
+
+  if (json.opponentRankingClass == "Mythic") {
+    const httpApi = require("../httpApi");
```
lusbenjamin: this can probably become a normal `import`

lusbenjamin:
still ~50 build errors in TS

src/window_background/onLabel/EventMatchCreated.ts
Manuel-777: This was because of circular dependencies, havent tried using an import after splitting the labels trough.

Manuel-777:
Took most of the ts errors down, I am having some issues with `InEventGetPlayerCourseV2` and `trackUpdate`.
The later is probably quite complex since it affects many more files, because of the way we minify data and expect it in the scripts.

Also, please be **SUPER** careful using this PR and backup your local data, since mose deck serialization code changed I am sure it will screw some data if used on its current state, my next goal is to fix that.

src/types/inventory.ts
```diff
@@ -59,7 +65,7 @@ export interface InternalEconomyTransaction {
   id: string;
   date: Date;
   context: string;
-  subContext: {
+  subContext?: {
```
Manuel-777: Is this only to fix an error or because you know `subContext` might be undefined some times?

src/window_background/onLabel/ClientToMatchServiceMessageTypeClientToGREMessage.ts
Manuel-777: This was part of a circular dependency issue, maaaaybe after separating everything will not be needed anymore and we can import normally.

src/types/inventory.ts
lusbenjamin: I thought that `subContext` was only used by some of the label handlers, but I did not double check to confirm that it was the case. (To the best of my memory, we invented `subContext` as an internal field to track some of the JSON that spawned multiple sub-JSON transactions. I don't _think_ they all behaved that way.)

lusbenjamin:
@Manuel-777 this is now caught up to the latest and greatest `develop`! I moved all of the `SerializedFoo` classes I had made during the `react-table` push into the common types area, as well as some of the original `ExtendedMatchData` types. I also renamed all of these types to match your current `InternalFoo` convention. Finally, I combined `CourseDeck` into `ArenaV3Deck` and refactored some of the shared deck-related properties between `ArenaV3Deck` and `InternalDeck`.

src/shared/deck.ts
```diff
@@ -39,16 +50,26 @@ class Deck {
...
+    this.id = mtgaDeck.id || "";
+
+    if (mtgaDeck.type !== "InternalDeck") {
+      this.tags = [mtgaDeck.format ?? "unknown"];
+    } else {
+      this.tags = [mtgaDeck.format ?? "unknown"] || mtgaDeck.tags;
+    }
+
+    this.custom =
+      mtgaDeck.type == "InternalDeck" ? mtgaDeck.custom || false : false;
```
lusbenjamin: I recently learned that we should use an explicit custom type predicate helper in these situations

lusbenjamin:
note that this now directly conflicts with #894, but this should be easier to resolve if we take care of #894 first and then resolve conflicts here later. Ideally, this PR should be left with just background-related refactors and a few upgrades to `shared` util classes.

src/window_background/addCustomDeck.ts
```diff
@@ -1,21 +1,14 @@
...
 const addCustomDeck = function(customDeck: Deck): void {
   const id = customDeck.id;
-  const deckData = {
-    // preserve custom fields if possible
-    ...(playerData.deck(id) || {}),
-    ...customDeck
-  };
-
+  const deckData = customDeck.getSave();
+  console.log("window_background addCustomDeck.ts deckData", deckData);
```
lusbenjamin: After extracting and testing #894, I think this change may be the one responsible for polluting @Manuel-777 's history with `Deck` objects. (It could also include the changeset in the calling locations.)

Manuel-777: I silently fixed that issue and it was related to this function indeed, the problem was originating in the `CardsList` class, where we accidentally passed another `CardsList` as a constructor argument.

lusbenjamin:
current plan is to revisit this after we tackle #916

lusbenjamin:
@Manuel-777 I finished auditing this PR. A quick summary:
- ensured that all changes outside of `window_background` were strictly lint or better than current `develop` (none should have any effect)
- ensured that all changes that have happened to code _INSIDE_ `window_background` from Dec 26 until now were ported forward into the corresponding label handlers in this PR (see all the "bugfix regression" commits)

Manuel-777:
Impressive reviewing work you have done!

I will do some more testing, but so far everything works as expected. The only thing I havent tested in depth was drafts, both overlays and replays. I will test those first.

PR #822, opened 2019-12-29
Decks table misc

lusbenjamin:
Fixes 2 misc regressions introduced in #814:
- restores the `cursor: pointer` style to the table rows and header cells
- bugfixes rendering performance by eliminating the react `key` error
  - instead of adding a `headerGroup` wrapping element with a key, I chose to simplify the number of rendering loops to make it unnecessary.

### DIFF
--- a/src/window_main/components/decks/DecksTable.tsx
+++ b/src/window_main/components/decks/DecksTable.tsx
@@ -455,29 +455,29 @@ export default function DecksTable({
       >
         {headers
           .filter((header: any) => header.isVisible)
-          .map((header: any, ii: number) => (
+          .map((column: any, ii: number) => (
             <div
-              {...header.getHeaderProps(header.getSortByToggleProps())}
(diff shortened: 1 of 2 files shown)


src/window_main/components/decks/DecksTable.tsx
```diff
@@ -452,88 +453,84 @@ export default function DecksTable({
         }}
         {...getTableProps()}
       >
-        {headerGroups.map((headerGroup: any) => {
-          return (
-            <>
```
lusbenjamin: This recently introduced fragment needs to have a `key` to allow React to render the table well for large numbers of rows. I chose to eliminate it instead. (We do not use grouping functionality, so we don't need `headerGroups` anyway.)

lusbenjamin:
note for reviewers: turn on "Ignore whitespace" and reload to make the diff nice

Manuel-777:
Looks good!

PR #827, opened 2019-12-31
Decks Table Control Upgrades

lusbenjamin:
### Motivation
This PR upgrades the Decks Table with a bunch of goodies:
- Moves the last chunk of state management logic on the decks tab into React
- `DecksTab` is now a proper React Component
- several performance optimizations (memoization and cleaner React hooks)
- connects the stats panel to the currently filtered decks in the table
- adds pagination controls to the decks table
- New dual view modes, "Table View" and "Deck Art View"

react-decks-tab

### DIFF
--- a/src/window_main/components/PagingControls.tsx
+++ b/src/window_main/components/PagingControls.tsx
@@ -31,30 +31,31 @@ export const StyledInputContainer = styled.div.attrs(props => ({
   }
 `;
 
-export default function PagingControls(props: {
-  canPreviousPage: any;
-  canNextPage: any;
+export interface PagingControlsProps {
+  canPreviousPage: boolean;
+  canNextPage: boolean;
   pageOptions: any;
-  pageCount: any;
+  pageCount: number;
   gotoPage: any;
   nextPage: any;
   previousPage: any;
   setPageSize: any;
-  pageIndex: any;
-  pageSize: any;
-}): JSX.Element {
-  const {
-    canPreviousPage,
-    canNextPage,
-    pageOptions,
-    pageCount,
-    gotoPage,
-    nextPage,
-    previousPage,
-    setPageSize,
-    pageIndex,
-    pageSize
-  } = props;
+  pageIndex: number;
+  pageSize: number;
+}
+
+export default function PagingControls({
+  canPreviousPage,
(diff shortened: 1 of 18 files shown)


Manuel-777:
Code looks so pretty now I cant find anything "strange" to look at.. mmmm

src/window_main/components/PagingControls.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <div className={"select_container"} style={{ width: "140px" }}>
+        <ReactSelect
+          current={String(pageSize)}
+          options={["10", "25", "50", "100"]}
+          optionFormatter={(pageSize): JSX.Element => (
+            <option key={pageSize} value={pageSize}>
+              Show {pageSize}
+            </option>
+          )}
+          callback={(val): void => setPageSize(Number(val))}
```
Manuel-777: Im not sure how its passed onto the tables, but this does not seem to be stateful, as changing one selector does not trigger the other selector using the same props to update.

src/window_main/components/decks/cells.tsx
```diff
@@ -354,7 +262,7 @@ export function MissingCardsCell({ cell }: CellProps): JSX.Element {
     mythic: pd.economy.wcMythic
   };
   return (
-    <StyledFlexRightCell>
+    <FlexLeftContainer>
```
Manuel-777: Not sure how to handle it, but the booster cost looks pretty bad in table view _(yes, im a big fan of table view!)_
Maybe it could look differently based on the view?  a simple text could work, of if possible an entirely different css layout that works in the reduced space

lusbenjamin:
Now includes all of the things I learned from building a second page with `react-table`!
- No more misc hard-coded lists of column ids scattered around: now uses custom column props
- No more horrid `CellWrapper` hack: now uses custom table instance props
- No more horrid `useControlledState` hook hack: no uses a simple state-driven `useEffect`
- Extracts common display components
  - all of the `styled` components now live in `display`
  - common `PagingControls` and `TableHeaders`
- g2g to reuse components from this for Collection Page V3!

Manuel-777:
Great stuff! merged so #829 can use its components now

PR #825, opened 2019-12-30
Collection V3

lusbenjamin:
### Motivation
Theros is almost here! Definitely the perfect excuse for me to give the Collection tab a little ❤️.

#### Code Health
- splits up `collection.js` into several smaller TypeScript modules, grouped together in `/collection`
- moves `collectionStats` into `/collection`
- enables more fancy spread syntax by adding `"downlevelIteration": true` to our tsconfig

#### Fancy New Stats Sidebar
Upgrade to the old "Statistics" page. Always respects the current search filters!
image

#### Promote Heatmap to Primary View
Upgrade to the old "Statistics" page. Always respects the current search filters!
image

#### Promote Set Statistics to Primary View
Upgrade to the old "Statistics" page. Always respects the current search filters!
image

#### More Goodies!
- Upgraded "Boosters Only" filter from "Statistics" page to an advanced filter. Affects all data views!
- Refactored "Export Collection" into "Export". Always respects the current search filters!
image

PR #829, opened 2020-01-02
Collection Page V3 w React Table

lusbenjamin:
### Motivation
Makes the Collection Page awesome. Stacked on #827, makes #825 obsolete.

react-collection-tab


lusbenjamin:
#### Feedback
- [ ] I see presets on top, maybe you a sets filter like we had before? I found it just perfect. Or just make the presets filter its own controls space so it stands out of the rest and is centered in the screen?
- [ ] I see the colors column, maybe we can add default "mana cost" column? it makes cards easier to look at on a table.
- [ ] the "owned" filter could easily be a 4 choice like the mana colors filter, we can make 4 new icons for that or just use the mana cost numbers 1, 2, 3 and 4, an input seems complicated to use

Manuel-777:
Made some regex wizardy to tokenize search data from the uber search box; it retrieves three pieces of data to help with filtering; token, separator and value.

It is smart enough to know when a value is quoted so queries like this should be absolutely possible;
`o:"enters the battlefield" cmc>=2 t:creature c=uw`

As long as we add the proper cases for each token (o, cmc, t, c, etcetera) and teach it to use the separators like `>`, `<`, `>=`, `<=` and `=`. Of course more separators are possible, its magic.

lusbenjamin:
This PR now also includes:
- bugfix `Colors` now correctly handles split-color mana cost symbols
- proper TS declarations from `@types/react-table` module
- extracted and generalized common table components and types in `components/tables`
  - shared by Collection and Decks tab, soon to be used by Economy tab, Matches tab, etc.
- minor bugfixes for the Decks table (lint, common component tweaks)

PR #831, opened 2020-01-07
[draft] Economy Page V3 w React Table

lusbenjamin:
stacked on #829 

uses the power of [`ReactTable.useGroupBy`](https://github.com/tannerlinsley/react-table/blob/master/docs/api/useGroupBy.md) to handle date-boundary grouping logic

image

image

PR #835, opened 2020-01-11
Multi React Table

lusbenjamin:
### Motivation
Converts the majority of the main window filtering and aggregation logic to using `react-table` hooks and components. Includes a bunch of bonus TypeScript conversion, too. (This was formerly #829, #831, #832, #834.)

### Common Code
- extracts and generalizes table components from Decks into `components/tables`
- adds proper types for `react-table` with `@types/react-table`
- removed deprecated filter logic from `Aggregator` and converted to TS
- deprecated and removed `FilterPanel`
- bugfix `Colors` now correctly handles split-color mana cost symbols
- `database.standardSetCodes` convenience getter
- extended `LocalTime` support for more `local-time` custom attributes and modes
- eslint enforces [react-hooks rules of hooks](https://reactjs.org/docs/hooks-rules.html)
- adds [react-autosuggest dependency](https://react-autosuggest.js.org/) and removes `autocomplete.js`
- tables (including the prior Decks page table) now handle too-many columns more gracefully

### Collection Page
- tons of new filters, sorts, and metrics available
- complete overhaul of "Collection Statistics" into several full data views
  - Cards View: like the v2 view
  - Table View: new react-table data view
  - Chart View: upgraded version of the chart that was in "Collection Statistics"
  - Sets View: upgraded version of the set-based progress view that was in "Collection Statistics"
- adds a collection stats draggable panel on the right
- adds "wanted" as blue pips to `OwnershipStars`, new metrics, and a preset

(Outdated demo)
react-collection-tab

### Economy Page
- tons of new filters, sorts, and metrics available
- uses the power of [`ReactTable.useGroupBy`](https://github.com/tannerlinsley/react-table/blob/master/docs/api/useGroupBy.md) to handle date-boundary grouping logic
- Bugfix: restores the sort-by-rarity on rewarded ICRs/event card pool transactions

image
image

### History Page
- Renames "History" to "Matches"
- tons of new filters, sorts, and metrics available
- removed Draft Replays (they no longer work at all on this page but are still on the Events page)

image
image

### Events Page
- tons of new filters, sorts, and metrics available
- adds a `StatsPanel` draggable panel on the right
- Draft events now show the rare/mythic picks (formerly on History page)

image

### Decks Page
- restores the event-based match data filter
- splits the the 4 wildcard metrics off of "Booster Cost" into independent columns
- bugfix: displays booster cost and wc metrics even when 0
- bugfix: displays all decks in the "Wanted" preset

image
image


lusbenjamin:
@Manuel-777: I cleaned up a ton of my copy-pasta from making 5 react-table pages. This now uses even more common components and several common hooks.

(PS: Please consider merging soon before this PR grows any bigger! 😉 )

src/shared/OwnershipStars.tsx
```diff
@@ -53,15 +70,19 @@ function MultiCardOwnership(props: { owned: number; acquired: number }) {
...
       ))}
     </>
   );
 }
 
-export default function OwnershipStars(props: { card: DbCardData }) {
-  const { card } = props;
+export default function OwnershipStars(props: {
+  card: DbCardData;
+  wanted?: number;
```
Thaoden: Instead of making this optional, I'd give it the default value `0`, saves you a bunch of `??` checks down the line.

src/shared/colors.ts
```diff
@@ -66,23 +66,25 @@ class Colors {
...
-        case "b":
-          this.b += 1;
-          break;
-        case "r":
-          this.r += 1;
-          break;
-        case "g":
-          this.g += 1;
-          break;
+    cost.forEach(symbol => {
```
Thaoden: Why the additional layer here?

src/shared/database.ts
```diff
@@ -239,6 +239,14 @@ class Database {
     return setCodes;
   }
 
+  get standardSetCodes(): string[] {
+    return this.sortedSetCodes.filter(
+      code =>
+        this.sets[code].collation !== -1 &&
+        new Date(this.sets[code].release) > new Date("2018-07-13")
```
Thaoden: What happened on 2018-07-13?

src/main.js
```diff
@@ -168,12 +168,7 @@ function startApp() {
 
   ipc.on("ipc_switch", function(event, method, from, arg, to) {
     if (debugIPC && method != "log_read") {
-      if (
-        debugIPC == 2 &&
-        method != "set_status" &&
-        method != "set_db" &&
-        method != "background_set_history_data"
-      ) {
+      if (debugIPC == 2 && method != "set_status" && method != "set_db") {
```
Thaoden: What's the meaning of the value "2"?

src/shared/database.ts
lusbenjamin: It simply represents a convenient cutoff date to filter out the sets that are legal in the current Standard format. I have extracted it to a named constant, but perhaps the more "correct" way to do this is in the metadata repo with a new property on sets?

src/shared/OwnershipStars.tsx
lusbenjamin: I left this property as optional to support current usage, but updated the rest of the file to expect a strict number.

src/shared/colors.ts
lusbenjamin: This is the bugfix that correctly handles dual-symbol mana costs, e.g. Deathless Knight, Burning-Tree Emissary, Arcanist's Owl, etc.

src/main.js
lusbenjamin: I don't know

src/shared/database.ts
Manuel-777: `<== PlayerInventory.GetFormats` contains data about the different formats, we can extract the sets from there and check against that.

Sounds like a separate PR work trough.

Thaoden: I'm fine with a named constant.

src/shared/OwnershipStars.tsx
```diff
@@ -14,36 +14,47 @@ function OwnershipInfinity(props: { owned: number; acquired: number }) {
...
+    color = "blue"; // not owned and wanted copy
+  }
   return <div className={`inventory_card_quantity_${color}`} title={title} />;
 }
 
-function MultiCardOwnership(props: { owned: number; acquired: number }) {
-  const { owned, acquired } = props;
+function MultiCardOwnership(props: OwnershipProps): JSX.Element {
+  const { owned, acquired, wanted } = props;
   let title = `${owned || 0}/4 copies in collection`;
```
Thaoden: `owned` is of type `number`, there shouldn't be any need to boolean-or it with 0 to get a value.

src/shared/OwnershipStars.tsx
```diff
@@ -14,36 +14,47 @@ function OwnershipInfinity(props: { owned: number; acquired: number }) {
...
   if (acquired) {
```
Thaoden:
```suggestion
if (acquired !== 0) {
```

src/shared/OwnershipStars.tsx
```diff
@@ -14,36 +14,47 @@ function OwnershipInfinity(props: { owned: number; acquired: number }) {
...
     title += ` (${acquired} recent)`;
   }
+  if (wanted) {
```
Thaoden:
```suggestion
  if (wanted !== 0) {
```

src/shared/OwnershipStars.tsx
```diff
@@ -53,24 +64,27 @@ function MultiCardOwnership(props: { owned: number; acquired: number }) {
...
+  const owned = playerData.cards.cards[card.id] ?? 0;
+  const acquired = playerData.cardsNew[card.id] ?? 0;
+  const wanted = props.wanted ?? 0;
   // TODO add custom logic to handle rats and petitioners
-  if (isbasic) {
-    return <OwnershipInfinity owned={owned} acquired={acquired} />;
-  }
-  return <MultiCardOwnership owned={owned} acquired={acquired} />;
+  const isbasic = cardHasType(card, "Basic Land");
+  const Renderer = isbasic ? OwnershipInfinity : MultiCardOwnership;
```
Thaoden: Nice one!
Just for a matter of taste, you could inline the `isbasic` one-liner assignment.

src/shared/OwnershipStars.tsx
```diff
@@ -4,8 +4,8 @@ import playerData from "./player-data";
 import { cardHasType } from "./cardTypes";
 import { DbCardData } from "./types/Metadata";
 
-function OwnershipInfinity(props: { owned: number; acquired: number }) {
-  const { owned, acquired } = props;
+function OwnershipInfinity(props: OwnershipProps): JSX.Element {
+  const { owned, acquired, wanted } = props;
   let title = "";
```
Thaoden:
```suggestion
  let title = (owned > 0 ? "∞" : "0") + " copies in collection";
```

Thaoden: Makes the next `if`-block obsolete.

src/main.js
lusbenjamin: Sorry, I realized that my answer was perhaps a little too terse! The change I made here was to remove the deprecated `"background_set_history_data"` from the expression, so I was preserving the other pre-existing logic without fully understanding it.

src/shared/OwnershipStars.tsx
```diff
@@ -14,36 +14,47 @@ function OwnershipInfinity(props: { owned: number; acquired: number }) {
...
 
-function OwnershipStar(props: OwnershipStarProps) {
-  const { owned, acquired, copyIndex, title } = props;
+function OwnershipStar(props: OwnershipStarProps): JSX.Element {
+  const { owned, acquired, wanted, copyIndex, title } = props;
   let color = "gray"; // default unowned
   if (copyIndex < owned) {
     color = "green"; // owned copy
   }
   if (acquired && copyIndex >= owned - acquired && copyIndex < owned) {
```
Thaoden:
```suggestion
  if (copyIndex >= owned - acquired && copyIndex < owned) {
```

Thaoden: In the degenerate case not explicitely checked anymore where `acquired === 0`, this would evaluate to `true && copyIndex >= owned && copyIndex < owned` which would then always evaluate to false.

src/shared/OwnershipStars.tsx
```diff
@@ -14,36 +14,47 @@ function OwnershipInfinity(props: { owned: number; acquired: number }) {
...
     color = "orange"; // owned and newly acquired copy
   }
+  if (wanted && copyIndex >= owned && copyIndex < owned + wanted) {
```
Thaoden:
```suggestion
  if (copyIndex >= owned && copyIndex < owned + wanted) {
```

Thaoden: By the same reasoning as above.

src/main.js
Thaoden: But maybe someone else knows what "2" stands for and we can name this magic number...

src/shared/OwnershipStars.tsx
Thaoden: The whole block could be rewritten as follows:
```
if (copyIndex < owned) {
  if (copyIndex >= owned - acquired) {
    color = "orange";
  } else {
    color = "green";
  } else if (copyIndex < owned + wanted) {
    color = "blue";
  } else {
  color = "gray";
  }
}
```

This could perhaps be moved into a helper method.

This whole construct could even be transformed into a nested ternary monstrosity but that's as usual a matter of taste.

src/shared/constants.ts
```diff
@@ -655,6 +655,7 @@ export const FORMATS: Record<string, string> = {
   GRN: "Ravnica Constructed",
   XLN: "Ixalan Constructed"
 };
+export const STANDARD_CUTOFF_DATE = "2018-07-13";
```
Thaoden: For later reference: This should probably be an item in the metadata database, given it updates regularly.

src/shared/time-components/LocalTime.tsx
```diff
@@ -3,17 +3,21 @@
 
 import React from "react";
 
+type numericOptions = "2-digit" | "numeric";
+type textOptions = "short" | "long";
```
Thaoden: What is the reasoning behind these new types?

src/window_background/background.js
```diff
@@ -237,8 +237,9 @@ ipc.on("save_user_settings", function(event, settings) {
   // console.log("save_user_settings");
   ipc_send("show_loading");
   let refresh = true;
-  if (settings.skip_refresh) {
+  if (settings.skip_refresh || settings.skipRefresh) {
```
Thaoden: What is the difference between `skip_refresh` and `skipRefresh`?

src/window_main/DateFilter.tsx
```diff
@@ -1,59 +1,83 @@
...
+const dateOptions = [
+  DATE_ALL_TIME,
+  DATE_SEASON,
+  DATE_LAST_30,
+  DATE_LAST_DAY,
+  "Custom"
+];
+
+function customDateFormatter(filter: Date | string): string {
+  if (dateOptions.includes(String(filter))) {
```
Thaoden: Not completely sure what `String()` does here...

src/window_main/DecksTab.tsx
```diff
@@ -110,31 +95,29 @@ function updateStatsPanel(
...
       const archivedSortVal = deck.archived ? 1 : deck.custom ? 0.5 : 0;
       const colorSortVal = deck.colors ? deck.colors.join("") : "";
       // compute winrate metrics
-      const deckStats: DeckStats =
-        aggregator.deckStats[id] ?? getDefaultStats();
-      const avgDuration = Math.round(deckStats.duration / deckStats.total);
-      const recentStats: DeckStats =
-        aggregator.deckRecentStats[id] ?? getDefaultStats();
+      const deckStats: AggregatorStats =
+        (aggregator.deckStats as any)[id] ?? Aggregator.getDefaultStats();
```
Thaoden: Downcast?

src/window_main/DecksTab.tsx
```diff
@@ -110,31 +95,29 @@ function updateStatsPanel(
   container.appendChild(statsPanelDiv);
 }
 
-function getDecksData(aggregator: any): DecksData[] {
+function getDecksData(aggregator: Aggregator): DecksData[] {
   return pd.deckList.map(
     (deck: SerializedDeck): DecksData => {
       const id = deck.id ?? "";
       const archivedSortVal = deck.archived ? 1 : deck.custom ? 0.5 : 0;
       const colorSortVal = deck.colors ? deck.colors.join("") : "";
```
Thaoden:
```suggestion
      const colorSortVal = deck.colors?.join("") ?? "";
```

src/shared/time-components/LocalTime.tsx
lusbenjamin: https://github.com/github/time-elements#options

src/window_background/background.js
lusbenjamin: they are identical parameters, this PR is adding the second one as an `eslint`-compatible future choice.

src/window_main/DecksTab.tsx
lusbenjamin: the first part of this PR was written before I converted `aggregator` to TS. I think this downcast is actually when I got fed up with it being a legacy file and started to convert it.

src/window_background/background.js
Thaoden: Is there a way to already deprecate one of them?

lusbenjamin: we could do a codebase search for `skip_refresh` and deprecate it entirely, yes.

src/window_main/EconomyTab.tsx
```diff
@@ -0,0 +1,123 @@
...
+  boosters.reduce(
+    (accumulator: number, booster: { count: number }) =>
+      accumulator + booster.count,
+    0
+  );
+
+function getTxnData(): TransactionData[] {
+  const today = new Date();
+  return pd.transactionList.map(
+    (txn: SerializedTransaction): TransactionData => {
```
Thaoden: Maybe move this fn into a helper method.

lusbenjamin: not trying to be stubborn, but why bother?

Thaoden: Readability :)

src/window_main/EventsTab.tsx
```diff
@@ -0,0 +1,252 @@
...
+    gameLosses: 0,
+    isMissingMatchData: false,
+    losses: 0,
+    matchIds: [],
+    wins: 0
+  };
+  if (
+    event.custom ||
+    event.CurrentEventState === "DoneWithMatches" ||
+    event.CurrentEventState === 2
```
Thaoden: What does "2" mean here?

src/window_main/EventsTab.tsx
```diff
@@ -0,0 +1,252 @@
...
+  ) {
+    stats.eventState = "Completed";
+  }
+  if (!eventData) {
```
Thaoden: `eventData` is assigned at line 80, so this is always `false`.

src/window_main/EventsTab.tsx
```diff
@@ -0,0 +1,252 @@
...
+    return stats;
+  }
+  if (eventData.ProcessedMatchIds) {
+    stats.matchIds = eventData.ProcessedMatchIds.map(getValidMatchId).filter(
+      id => id
```
Thaoden: What does the filter achieve?

lusbenjamin: I was not the original author, but I think the general idea is to use the list of match IDs that Arena says were part of the event if possible. In the cases where mtgatool failed to record those matches, the page uses different logic to display the stats for the event?

Thaoden: I thought that would `getValidMatchId` do already, but I don't really know it either.

src/window_background/background.js
Thaoden: Want to do it here or in its own PR?

src/window_main/components/decks/DecksArtViewRow.tsx
```diff
@@ -0,0 +1,66 @@
...
+      onMouseEnter={mouseEnter}
+      onMouseLeave={mouseLeave}
+      {...divProps}
+    >
+      <CSSTransition classNames="deckTileHover" in={!!hover} timeout={200}>
+        <DeckArt url={getCardArtCrop(row.values["deckTileId"])} />
+      </CSSTransition>
+      {row.cells.map(cell => {
+        // cell.hover = hover;
+        return (
```
Thaoden: For readability, I'd extract this `return` into a helper.

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+    targets: ".moving_ux",
+    left: "-100%",
+    easing: EASING_DEFAULT,
+    duration: 350
+  });
+}
+
+function addTag(matchid: string, tag: string): void {
+  const match = pd.match(matchid);
+  if (!match || !tag) return;
```
Thaoden: Both `match` and `tag` are typed as `string`, this line will never evaluate to `true`.

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+  if ([tagPrompt, NO_ARCH, DEFAULT_ARCH].includes(tag)) return;
+  if (match.tags && match.tags.includes(tag)) return;
```
Thaoden:
```suggestion
 if (match.tags?.includes(tag)) return;
```

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+  ipcSend("add_matches_tag", { matchid, tag });
+}
+
+function editTag(tag: string, color: string): void {
+  ipcSend("edit_tag", { tag, color });
+}
+
+function deleteTag(matchid: string, tag: string): void {
+  const match = pd.match(matchid);
+  if (!match || !tag) return;
```
Thaoden: This `if` always evaluates to `false`.

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+  if (!match.tags || !match.tags.includes(tag)) return;
```
Thaoden:
```suggestion
 if (!match.tags?.includes(tag)) return;
```

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+  ipcSend("delete_matches_tag", { matchid, tag });
+}
+
+function getNextRank(currentRank: string): undefined | string {
+  const rankIndex = (RANKS as any).indexOf(currentRank);
```
Thaoden: Cast to `any`?

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+  if (rankIndex < RANKS.length - 1) {
+    return RANKS[rankIndex + 1];
+  } else {
```
Thaoden: nit: `else` is redundant here, you can just `return undefined` outside of the `if`-block.

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+}
+
+function getStepsUntilNextRank(mode: boolean, winrate: number): string {
+  const rr = mode ? pd.rank.limited : pd.rank.constructed;
+
+  const cr = rr.rank;
+  const cs = rr.step;
+  const ct = rr.tier;
+
+  let st = 1;
```
Thaoden: For later reference: These variables (constants, really) could be moved into `constants` and maybe even put into a dictionary keyed by `cr`.

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  const expectedValue = winrate * stw - (1 - winrate) * stl;
+  if (expectedValue <= 0) return "&#x221e";
+
+  const stepsNeeded = st * ct - cs;
+  let expected = 0;
+  let n = 0;
+  // console.log("stepsNeeded", stepsNeeded);
+  while (expected <= stepsNeeded) {
```
Thaoden: Isn't this just calculating `n = stepsNeeded % expectedValue`?

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+  return "~" + n;
+}
+
+function renderRanksStats(
+  container: HTMLElement,
+  aggregator: Aggregator,
+  isLimited: boolean
+): void {
+  container.innerHTML = "";
+  if (!aggregator || !aggregator.stats?.total) return;
```
Thaoden: `aggregator` should not happen to be `undefined | null` here according to its type, so the first part should never evaluate to `true`.

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+  div.style.padding = "0 12px";
+
+  if (showingRanked) {
+    const rankStats = createDiv(["ranks_stats"]);
+    renderRanksStats(rankStats, aggregator, isLimited);
+    rankStats.style.paddingBottom = "16px";
+    div.appendChild(rankStats);
+
+    rankedStats =
+      eventId === RANKED_CONST
```
Thaoden: This check is made redundant by line 200. You can just set `rankedStats = aggregator.limitedStats` here and line 200 will take care of setting it so `constructedStats` if need be.

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+        return false;
+      }
+      return aggregator.filterMatch(match);
+    })
+    .map(
+      (match: SerializedMatch): MatchTableData => {
+        const timestamp = new Date(match.date ?? NaN);
+        const colors = match.playerDeck.colors ?? [];
+        const oppColors = match.oppDeck.colors ?? [];
+        return {
```
Thaoden: I feel like I've encountered some of these object creations in various files (can't really check though). If so, maybe extract this into a helper function somewhere.
Even if not, maybe extract it to improve readability (or maybe even make `MatchTableData` a class caring about the checks itself?)

src/window_main/MatchesTab.tsx
```diff
@@ -0,0 +1,367 @@
...
+function getTotalAggData(): [string[], TagCounts] {
+  const totalAgg = new Aggregator();
+  const allTags = [
+    ...(totalAgg.archs ?? []).filter(
+      arch => arch !== NO_ARCH && arch !== DEFAULT_ARCH
+    ),
+    ...Object.values(db.archetypes).map(arch => arch.name)
+  ];
+  const tags = [...new Set(allTags)].map(tag => {
+    const count = totalAgg.archCounts?.[String(tag)] ?? 0;
```
Thaoden: What does `String(tag)` do here?

src/window_main/EventsTab.tsx
lusbenjamin: Now I understand the question. The final call to `filter` in this call chain filters out the `undefined` results that could potentially return from `getValidMatchId`. I added explicit syntax to clarify.

src/window_main/MatchesTab.tsx
lusbenjamin: required to allow TS to use `indexOf`

src/window_main/MatchesTab.tsx
lusbenjamin: It looks like it is doing a `Math.ceil(stepsNeeded / expectedValue)`, yes. Refactoring to simplify.

src/window_main/MatchesTab.tsx
lusbenjamin: this logic is odd, refactoring to isolate it and attempt to make it more explicit

src/window_main/DateFilter.tsx
lusbenjamin: Refactoring to do the same logic with explicit `typeof`

src/window_main/MatchesTab.tsx
lusbenjamin: it is logcially equivalent to `[tag]`, which TS rejects. Another possible syntax that TS allows is `[tag ?? ""]`.

src/window_background/background.js
lusbenjamin: separate PR sounds better

src/window_main/aggregator.ts
```diff
@@ -6,114 +6,117 @@ import max from "date-fns/max";
...
       (match.eventId && match.eventId.includes("Draft")) ||
       (match.type && match.type === "draft")
     );
   }
 
-  static gatherTags(_decks) {
-    const tagSet = new Set();
-    const formatSet = new Set();
-    const counts = {};
+  private static gatherTags(_decks: any[]): string[] {
```
Thaoden: nit: `decks` instead of `_decks`

src/window_main/aggregator.ts
```diff
@@ -144,113 +181,63 @@ class Aggregator {
...
-    const passesColorFilter = Aggregator.filterDeckByColors(deck, colors);
-    if (!passesColorFilter) return false;
-
-    return true;
+      this.validDecks.has(deck.id)
+    );
   }
 
-  filterEvent(_eventId) {
+  filterEvent(_eventId: string): boolean {
```
Thaoden: nit: `eventId` instead of `_eventId` (or something else to prevent conflicts with line 200).

src/window_main/aggregator.ts
```diff
@@ -144,113 +181,63 @@ class Aggregator {
...
       eventId === _eventId
     );
   }
 
-  filterMatch(match) {
+  filterMatch(match: any): boolean {
     if (!match) return false;
-    const { eventId, oppColors, arch, showArchived } = this.filters;
+    const { showArchived, matchIds } = this.filters;
     if (!showArchived && match.archived && match.archived) return false;
```
Thaoden: nit: redundant `&& match.archived`

src/window_main/aggregator.ts
```diff
@@ -383,7 +369,9 @@ class Aggregator {
       }
       // process archetype
       const tag =
-        match.tags && match.tags.length ? match.tags[0] || NO_ARCH : NO_ARCH;
+        match.tags && match.tags.length
```
Thaoden:
```suggestion
     match.tags?.length !== 0
```

src/window_main/aggregator.ts
```diff
@@ -383,7 +369,9 @@ class Aggregator {
...
+          ? match.tags[0] || Aggregator.NO_ARCH
+          : Aggregator.NO_ARCH;
       this.archCounts[tag] = (this.archCounts[tag] || 0) + 1;
```
Thaoden:
```suggestion
      this.archCounts[tag] = (this.archCounts[tag] ?? 0) + 1;
```

src/window_main/collection/CollectionTab.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+      add = add.replace(
+        "$Count",
+        pd.cards.cards[key] === 9999 ? 1 : pd.cards.cards[key]
+      );
+
+      add = add.replace("$SetName", card.set);
+      if (card.set in db.sets)
+        add = add.replace("$SetCode", db.sets[card.set].code);
+      add = add.replace("$Collector", card.cid);
```
Thaoden: nit: these `replace`s can all be chained together in a single call.

lusbenjamin:
@Manuel-777 this now attempts to handle the overflow-x and overflow-y the way we discussed, but it needs help! I couldn't figure out how to get the contents of the table to correctly fill out into the hidden scroll space. Can you assist with your CSS-fu?

PR #838, opened 2020-01-11
bugfix old Deck data may be Arena Deck v3

lusbenjamin:
### Motivation
Some time ago, there existed some released Beta or stable release that allowed users to save invalid draft decks. This was likely immediately after the log format changed to Arena V3, but before we correctly updated our label handlers. As a result, @Thaoden had several draft decks with an invalid format that were causing bugs on the Decks page and the Collection page.

This PR adds a step to our initial tool startup to automatically detect and convert those Arena V3 decks into the correct mtgatool internal format.

### DIFF
--- a/src/shared/util.js
+++ b/src/shared/util.js
@@ -320,10 +320,7 @@ export function get_wc_missing(deck, grpid, isSideboard) {
   // cap at 4 copies to handle petitioners, rat colony, etc
   needed = Math.min(4, needed);
 
-  const card = db.card(grpid);
-  if (!card) {
-    return 0;
-  }
+  let card = db.card(grpid);
   let arr = card.reprints;
   if (!arr) arr = [grpid];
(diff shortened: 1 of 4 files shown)


Thaoden:
Maybe instead of checking and returning `0`, you could check the card and assign `arr` to an empty array, that way you can use the remaining existing logic instead of short-cutting.

src/shared/util.js
```diff
@@ -320,7 +320,10 @@ export function get_wc_missing(deck, grpid, isSideboard) {
   // cap at 4 copies to handle petitioners, rat colony, etc
   needed = Math.min(4, needed);
 
-  let card = db.card(grpid);
+  const card = db.card(grpid);
+  if (!card) {
```
Thaoden: This guard should no longer be necessary.

Issue #581, opened 2019-09-13
Users should be able to seamlessly switch between Arena accounts


AnnanFay:
Most of this is transcribed from [this discord discussion](https://discordapp.com/channels/463844727654187020/506793351786266624/622030645862072326).

**Users should be able to switch between Arena accounts without data being lost or merged together.**

The likely use case is someone who is encouraging their SO or sibling into playing the game while having the tracker running in the background. Second likely use case is a streamer running a F2P account.

Currently there is at least one issue with decklists being merged, and a few things untested. I've not even looked into how the sync behaves because I know next to nothing about it.

Here is a test log file of logging into an account, immediately logging out then logging into a completely brand new account. Detailed logging enabled throughout. New user account, tutorial games and 1 play game. Creating new deck, progression along "Account Mastery", etc. Logout and login happens at line 62094

The likely expected behaviour of arena tool by users depends when arena tool was started.
If Arena Tool was started after the second account logged in then the AT account and data should be for the second account only.
If AT was started while the first account was logged in then the AT account should be for the first arena user.

When loading the above log file into a fresh account in offline mode the following was observed:

* Second account detected (correct behaviour)
* Decks from both accounts merged
* Quite a economy events are missing pretty text

![image](https://user-images.githubusercontent.com/766048/64861479-f42a6b00-d627-11e9-9eb3-0c67efba4cee.png)

![image](https://user-images.githubusercontent.com/766048/64861464-ebd23000-d627-11e9-8c6e-0cb297632bef.png)

This is a limited test because no matches were done on the first account and it wasn't done while the tracker was active.

We probably should display the tutorial games and bot games for new users. Right now we are completely avoiding them, but a better option would be to show them for new users - account level <25 and to auto-archive them for experienced users.

Stuff to test:

* Play matches in both accounts.
* Test syncing.
* Running test while tracker is active and logged into first account.
* Multiple account switches per log file.

PR #840, opened 2020-01-13
fix multi account merges

Manuel-777:
Smoothly switch between players data based on the connection events from the logs.

Still havent enabled any way to force load a player's data (@AnnanFay suggested this at some point), and while its possible I believe it may be better to only allow that if the login in MTG Arena actually happened, by reading the connected event (like this PR does), that is to only allow local users looking at the data from the account only if they actually logged into it.
Also doing this kind of unexpected "switch" while another user is playing might cause the merging issues we want to avoid so badly in the first place unless we implement some complex logic of "this user is being used but data displayed if from this other user".

### DIFF
--- a/src/window_background/background.js
+++ b/src/window_background/background.js
@@ -504,7 +504,7 @@ async function logLoop() {
 
     // Get player Id
     strCheck = '\\"playerId\\": \\"';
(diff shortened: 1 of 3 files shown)


AnnanFay:
Related #581

PR #842, opened 2020-01-17
Revert "attempt to fix multi account merges (#840)"

lusbenjamin:
Currently `develop` has a broken build caused by #840. This simply reverts that commit.

(most likely the root problem is a circular dependency issue like `labels --> loadPlayerConfig --> arena-log-watcher --> labels --> loadPlayerConfig ...`)

This reverts commit 32e709597d309fb5722abc158def478f5785a5e8.


lusbenjamin:
closing in favor of #843

PR #843, opened 2020-01-17
Fix loadPlayerConfig path

Thaoden:
`loadPlayerConfig` is in the same folder as `labels`, so only a single dot necessary.


Manuel-777:
Told you this was working lol

PR #846, opened 2020-01-18
release fixes

Manuel-777:


lusbenjamin:
crazy looking commit history even though the actual delta is small. we probably need to either merge `master` into `develop` or spawn a new `develop` off of current `master`

Manuel-777:
There goes one thing I dont quite understand about git heh.. the commits are in both branches so there is no need to include them in the history. Probably the best way to do this is cherry picking the commits when creating the release PRs ?

PR #847, opened 2020-01-18
Sync Commit History

lusbenjamin:
### Motivation
`master` is functionally caught up to `develop` but appears to be behind due to our use of squash-merge commits. Merging this PR as a normal merge should bring the commit histories more correctly together again.


lusbenjamin:
we should ship a performance hotfix ASAP, so this PR is redundant

PR #848, opened 2020-01-18
Hotfix login performance

lusbenjamin:
### Motivation
Current stable 3.0.2 is guaranteed to hit a worst-case performance case during initial boot. The first pass of the log (not the initial scan to detect user) will find the log statement where the user logged in to Arena and spawn a redundant `loadUserConfig` call. This call is asynchronous, extremely expensive, and creates a worst-case fight over resources with the pre-existing first pass of the log.

To solve this, we simply have to revert #840.

NOTE: please do a normal merge instead of a squash to ensure that future PRs to `master` have a clean commit history.


src/window_background/labels.js
```diff
@@ -298,12 +297,6 @@ export function onLabelOutLogInfo(entry) {
 
   if (json.params.messageName == "Client.Connected") {
     logLanguage = json.params.payloadObject.settings.language.language;
-    const parsedData = {};
-    parsedData.arenaId = json.params.payloadObject.playerId;
-    parsedData.name = json.params.payloadObject.screenName;
-    parsedData.arenaVersion = json.params.payloadObject.clientVersion;
-    setData(parsedData, false);
-    loadPlayerConfig(json.params.payloadObject.playerId);
```
lusbenjamin: In our future fix for the multi-user case, we cannot simply call `loadPlayerConifg` like a standard label handler because it is an `async` function

PR #860, opened 2020-01-24
Dynamic React ListItem components

lusbenjamin:
### Motivation
This PR better supports the popular "List View" table displays by upgrading the legacy renderers into proper React components. This also takes care of several hundred lines of the TODO code in `renderer-util.js`.

BONUS: List Views now all respond to the column toggles and support displaying new metrics!
image


src/window_main/components/ListItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+function ItemSection({
+  section,
+  className
+}: {
+  section: ListItemSection;
+  className: string;
+}): JSX.Element {
+  if (
```
Thaoden: For my understanding: The `if` is checking the type of `section` to determine if it is a `SplitNode`?

lusbenjamin: Yes, but I couldn't figure out a good way to do it. Any suggestions?

src/window_main/components/decks/DecksListViewRow.tsx
```diff
@@ -1,116 +1,213 @@
...
+  addTagCallback,
+  editTagCallback,
+  deleteTagCallback
+}: DeckListItemProps): JSX.Element {
+  const grpId = deck.deckTileId ?? DEFAULT_TILE;
   const parentId = deck.id ?? "";
+  const onClick = (): void => openDeckCallback(parentId);
+  const onClickDelete = deck.custom
+    ? (): void => archiveCallback(parentId)
+    : undefined;
```
Thaoden: Better to use a no-op function here instead of setting it to `undefined`: `() => {}`

src/window_main/components/decks/DecksListViewRow.tsx
```diff
@@ -1,116 +1,213 @@
...
-    openDeckCallback,
-    archiveCallback,
-    addTagCallback,
-    editTagCallback,
-    deleteTagCallback
-  );
-  return <div title={"show deck details"} ref={containerRef} />;
+  // RIGHT SECTION
+  const missingCards = deck.common + deck.uncommon + deck.rare + deck.mythic;
+  let right;
```
Thaoden: nit: you could assign a type.

src/window_main/components/decks/DecksListViewRow.tsx
```diff
@@ -1,116 +1,213 @@
...
+      rare: pd.economy.wcRare,
+      mythic: pd.economy.wcMythic
+    };
+    right = (
+      <>
+        {CARD_RARITIES.filter(rarity => rarity !== "land").map(rarity => {
+          const cardRarity = rarity.toLowerCase();
+          const missing =
+            ((deck as unknown) as Record<string, number>)[cardRarity] ?? 0;
+          if (missing) {
```
Thaoden:
```suggestion
          if (missing !== 0) {
```

src/window_main/components/decks/DecksListViewRow.tsx
```diff
@@ -1,116 +1,213 @@
...
+      tooltip = formatWinrateInterval(
+        formatPercent(deck.winrateLow),
+        formatPercent(deck.winrateHigh)
+      );
+    } else {
+      // sample size is too small (garbage results)
+      interval = "???";
+      tooltip = "play at least 20 matches to estimate actual winrate";
+    }
+    right = {
```
Thaoden: Can the creation of `right` be extracted into a helper function?

src/window_main/components/events/EventsListViewRow.tsx
```diff
@@ -42,37 +37,61 @@ function handleOpenDraft(id: string | number): void {
...
-        ? ["list_event_phase"]
-        : ["list_event_phase_red"],
-      eventState
+  const grpId = event.CourseDeck.deckTileId;
+  const parentId = event.id;
+  const [expanded, setExpanded] = React.useState(false);
+  const onClick = (): void => setExpanded(!expanded);
+  const onClickDelete = event.custom
+    ? (): void => toggleArchived(parentId)
+    : undefined;
```
Thaoden: Maybe use a no-op function.

src/window_main/components/events/EventsListViewRow.tsx
```diff
@@ -42,37 +37,61 @@ function handleOpenDraft(id: string | number): void {
...
-        toMMSS(duration ?? 0) +
-        " long"
-    )
-  );
+  // CENTER SECTION
+  let center;
+  const draftId = event.id + "-draft";
+  if (pd.draftExists(draftId)) {
+    const draft = pd.draft(draftId);
+    const highlightCards: DbCardData[] = draft.pickedCards
```
Thaoden: This code is duplicated in `DraftListItem.tsx`. Maybe consider a utility function?

src/window_main/components/matches/MatchesListViewRow.tsx
```diff
@@ -1,110 +1,143 @@
...
-    deleteTagCallback
-  );
-  return <div title={"show match details"} ref={containerRef} />;
+}: MatchListItemProps): JSX.Element {
+  const grpId = match.playerDeck.deckTileId ?? DEFAULT_TILE;
+  const parentId = match.id;
+  const onClick = (): void => openMatchCallback(parentId);
+  const onClickDelete = archiveCallback
+    ? (): void => archiveCallback(parentId)
+    : undefined;
```
Thaoden: Consider a no-op function.

src/window_main/renderer-util.js
```diff
@@ -847,266 +847,12 @@ function localTimeSince(date) {
...
-}
-
-function draftShareLink(id, draft) {
-  const shareExpire = byId("expire_select").value;
+export function draftShareLink(id, draft) {
+  const shareExpire = byId("expire_select")?.value ?? "One day";
   const draftData = JSON.stringify(draft);
   let expire = 0;
   switch (shareExpire) {
+    default:
```
Thaoden: I _think_ this will prevent `"Never"` from being hit as cases are evaluated in a top-down manner and `default` short-circuits evaluation.

src/window_main/components/ListItem.tsx
Thaoden: Have a look at [User Defined Type Guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) from the TS docs. Basically, you introduce a helper function that checks the given object for presence or absence of a property that only exsist on one of the unioned types.

src/window_main/renderer-util.js
lusbenjamin: Good catch! Might have been an accidental merge or copy-paste bug?

Thaoden: What you could do here is not handle `One day` explicitly and let it fall through. Though I think in this case it is more clear when it is more explicit.

src/window_main/components/decks/DecksListViewRow.tsx
lusbenjamin: I slightly prefer the convention of matching the type signature of the destination property. For an optional callback handler property `onDelete?: () => void`, it feels more correct to use `undefined` and match the destination type.

src/window_main/components/decks/DecksListViewRow.tsx
lusbenjamin: Each subsection of each list item could be broken into a separate Component, similar to the structure that we see in the current Economy components. That level of fragmentation starts to make the code harder for me to read, so I chose not to do it initially.

I could definitely be convinced to create a helper component for particularly complex subsections, like this `right` subsection you have highlighted here. Refactoring this subsection would not make the code any easier to maintain from a logical organization perspective (the right subsection depends on almost all of the local state), but could have a marginal performance gain.

Manuel-777: Hold my beer...

src/window_main/components/ListItem.tsx
lusbenjamin: Great tip, thanks!

src/window_main/components/decks/DecksListViewRow.tsx
Thaoden: Fair enough.

src/window_main/components/ListItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+function ItemSection({
+  section,
+  className
+}: {
+  section: ListItemSection;
+  className: string;
+}): JSX.Element {
+  if (isSplitNode(section)) {
```
Thaoden: nit: you could use a `return isSplitNode ? : ` ternary here.

src/window_main/components/decks/DecksListViewRow.tsx
Thaoden: Personally I find extracting big chunks of assignments into a helper function to increase readability (`right = createWinrateStatsElement(bunch_of_data)`, `right = createDeckDraftingCostElement(bunch_of_other_data) `).

lusbenjamin:
This PR now has DYNAMIC List Views! Toggle some columns and watch the madness!

lusbenjamin:
closing this in favor of static list items in #862

PR #862, opened 2020-01-24
ListItem component

Manuel-777:
Draft PR


src/window_main/components/ManaCost.tsx
```diff
@@ -0,0 +1,28 @@
...
+interface ManaCostProps {
+  colors: number[];
+  class?: string;
+}
+
+export default function ManaCost(props: ManaCostProps): JSX.Element {
+  const { colors } = props;
+  // Default to size 16px, Initially these had classes because "s" was for
+  // shadowed mana costs, whereas no prefix was regular, non shadowed icon.
+  // I supose these could be a set of props instead.
```
lusbenjamin: so that's what the "s" means!

src/window_main/components/ListItem.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+interface FlexProps extends JSX.ElementChildrenAttribute {
+  title?: string;
+  innerClass?: string;
+}
+
+// These two are technically equal, but I like the way
+// of using them as different components, it makes coding
+// complex list items prettier.
+export function FlexTop(props: FlexProps): JSX.Element {
```
lusbenjamin: If they are most likely going to stay equal, you could do something like:
```js
export const FlexBottom = FlexTop;
```

Manuel-777:
I feel this is almost ready, I just need to bind the share buttons to actual share dialog and remove all unused codes left around.

PR #737, opened 2019-11-19
Webpack

lusbenjamin:
Moves our build process from directly invoking `babel` to using `webpack`. This should probably be a breaking change aimed at v3.0.

Resources:
- https://webpack.js.org/configuration/
- https://github.com/Devtography/electron-react-typescript-webpack-boilerplate
- https://www.codementor.io/randyfindley/how-to-build-an-electron-app-using-create-react-app-and-electron-builder-ss1k0sfer

### Done
- Build all of the TS
- move static assets into webpack
- Get all of the imports to work
- Keep `electron-builder` working

### Future Optimizations
- Only ship the contents of `lib` (requires file reorg?)
- hook up a local server and try hot reload hotness

### TODO
- [ ] more manual testing of dev mode
- [ ] test distribution end-to-end
- [ ] fix all dynamic references to static assets, for example the land icon in the overlay is currently broken for the "compact lands" feature due to this: https://github.com/Manuel-777/MTG-Arena-Tool/blob/master/src/shared/CardTile.tsx#L349

PR #863, opened 2020-01-27
No file extensions

lusbenjamin:
### Motivation
This PR completes a minor code chore to remove the few remaining explicit filename extensions from `import` statements. It is a pre-requisite for the long-delayed conversion of the build to Webpack (#737) and original came from that PR.


Manuel-777:
Looks good! Although I can't check now if those are all

PR #871, opened 2020-01-29
Bugfix remember previous scroll position

lusbenjamin:
### Motivation
In our legacy code (still includes the Explore page), we used `DataScroller` and `lastScrollTop` in `localState` to remember the user's current scroll position. If they chose to edit a row (add tag, delete tag, archive, etc), the round-trip to the background process would trigger a screen "refresh". At that point, `DataScroller` handled automatically restoring the user to their previous scroll position.

In the post-react-table world, we still have to do a round-trip in some cases, but we no longer have `DataScroller`. Hence this regression:
https://discordapp.com/channels/463844727654187020/467737642306371584/671912594558287872
image
^ this problem is broader than described here; it also affects the Decks page, Events page, and Economy page; it also affects archiving actions.

### Approach
This PR creates a new common hook that imitates the old approach in `DataScroller`, then adds the hook to the new react-table pages.

### DIFF
--- a/src/window_main/DecksTab.tsx
+++ b/src/window_main/DecksTab.tsx
@@ -18,7 +18,10 @@ import Aggregator, {
 } from "./aggregator";
 import DecksTable from "./components/decks/DecksTable";
 import { DecksData } from "./components/decks/types";
-import { useAggregatorAndSidePanel } from "./components/tables/hooks";
+import {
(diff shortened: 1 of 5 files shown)


Manuel-777:
Oops, this one should have been merged in v3.0.5 !

PR #872, opened 2020-01-29
Misc table tweaks

lusbenjamin:
This PR has a handful of low-risk tweaks to the table displays and types that were originally part of #860:
- cleans up some random code cruft
- breaks apart opponent Arena ID and name on matches page
- adds a new `SubTextCell` (italic) and uses it for some columns to match list items
- slightly more precise types


lusbenjamin:
Pretty sure this build passes, but I've never seen this type of Travis error before. Couldn't figure out how to restart the job.

PR #875, opened 2020-01-30
[WIP] React main window controller

lusbenjamin:
### Motivation
Converts the main window renderer and state into React components and hooks. Should ideally be at exact parity and have no noticeable effect to the user.


Manuel-777:
We have a ton of mounts and unmounts currently, I'd wait until all of the other renderer components are React to try this properly, not that the code is incorrect but because we might change the approach and implementation after it's all good.

I'd leave it open anyway, FFR

lusbenjamin:
> "We have a ton of mounts and unmounts currently"

I agree! This is one of my motivations for doing this, and my original plan was only to turn this into a PR after I had deprecated `mountReactComponent`. Since this PR gives control of the DOM "root" to React, all DOM children including `ux0` can theoretically start to depend on built-in React garbage collection. Said another way, my hope is that after this PR we only have to support React-wrapping-legacy-code and no longer Legacy-code-wrapping-React.

PR #880, opened 2020-02-02
[feature] draft set-completion estimation stats

lusbenjamin:
### Motivation
F2P and frugal players have frequently requested a metric to help them estimate "How many drafts will it take to get to the center of a Tootsie Pop?" The most popular approach so far was written up here: https://www.mtggoldfish.com/articles/collecting-mtg-arena-part-1-of-2 and has been exactly reproduced here: https://mtgahelper.com/draftsBeforeBoosters

As a quick rehash, I still think the currently accepted approach is wrong for these reasons:
  1. Major: ignores opportunity cost of hoarding boosters (assumes user does not play constructred events)
  2. Major: ignores law of diminishing marginal returns (draft estimation numbers are bad ceilings)
  3. ~~Minor: gets the probability of wildcard upgrade incorrect (uses 11/12 instead of 23/24)~~

Nevertheless, I want to build that approach exactly as a first iteration, then improve upon it after we can validate the numbers

### Initial Demo
image

### DIFF
--- a/src/window_main/collection/CollectionTab.tsx
+++ b/src/window_main/collection/CollectionTab.tsx
@@ -92,9 +92,12 @@ function isBoosterMathValid(filters: Filters<CardsData>): boolean {
   for (const filter of filters) {
     if (filter.id === "booster") {
       hasCorrectBoosterFilter = filter.value?.true && !filter.value?.false;
-    }
-    if (filter.id === "rarity") {
+    } else if (filter.id === "rarity") {
       hasCorrectRarityFilter = filter.value?.mythic && filter.value?.rare;
+    } else if (filter.id === "set") {
+      continue; // this is fine
+    } else {
+      return false; // no other filters allowed
     }
   }
   return hasCorrectBoosterFilter && hasCorrectRarityFilter;
--- a/src/window_main/collection/CollectionTab.tsx
+++ b/src/window_main/collection/CollectionTab.tsx
@@ -20,7 +20,8 @@ import {
   hideLoadingBars,
   ipcSend,
   makeResizable,
-  resetMainContainer
+  resetMainContainer,
+  setLocalState
 } from "../renderer-util";
 import {
   CollectionStats,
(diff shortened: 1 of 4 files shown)


dougmill:
> Minor: gets the probability of wildcard upgrade incorrect (uses 11/12 instead of 23/24)
If you're talking about the odds of specifically a rare WC, or specifically a mythic WC, then you are correct. If you're talking about the odds of getting _any_ wildcard in place of the rare/mythic, then the current code is correct. There is a 1 in 24 chance of getting a rare WC, and also a 1 in 24 chance of getting a mythic WC. The chance of getting a rare _or_ mythic WC is thus 2/24, or 1/12.

lusbenjamin:
The error in the OP math has to do with how the factor is applied in the total estimation. Each rarity is handled as an independent event and thus must NOT use the combined probability of a WC being upgraded.

dougmill:
Having now checked the linked article, I have verified that 11/12 is the correct number.

The chance of a mythic WC is not the fraction of mythics that are wildcards, it's the fraction of _packs_ that contain a mythic WC. Likewise for the rare WC chance. The formulas in question are for the chance of getting a non-wildcard rare, and the chance of getting a non-wildcard mythic. For _either_ of those outcomes to happen, the card has to not be _any_ kind of wildcard, so each formula has to account for both rare and mythic wildcards combined.

lusbenjamin:
I think I understand now, and I agree that OP math is correct. Just to confirm, the WC drop rate is the lower model pictured here (the top was my incorrect original model):
image

dougmill:
Confirmed, the lower model in the image is the correct one, and has the correct numbers.

lusbenjamin:
PS - after building this, I definitely regret not just doing this earlier. This one goes out with special thanks to all the users that kept asking for this and arguing in favor of it: @barduck, @Balefirex, @Infusion, @Lego6245, @Bleetz, and others I'm sure I've forgotten to mention.

lusbenjamin:
Since all of the booster-related math only makes sense when thinking about cards-available-in-boosters, this PR also now does the following:
- creates a "Boosters" preset button
- adds logic to detect when the current filters have a valid domain for booster-related math
- only shows booster-related math and inputs when filters have a relevant domain
- when filters do not have a relevant domain, shows a help message

image

Manuel-777:
Would it be possible to show the statistics without having turn the preset, as a default of the sets view? Maybe add a little notice saying it doesn't account for filters

PR #883, opened 2020-02-03
support wed brawl weekly events

lusbenjamin:
### Motivation
Every Wednesday, WotC celebrates its most controversial Arena event format by letting folks play Brawl for 24 hours. They implement these "Play Brawl" queues with a separate individual event each week. As a result, our event metadata repo struggles to keep up with adding new codes and mappings. This PR supports Wednesday Play Brawl queues by encapsulating the pattern WotC uses to generate the event code each week and then autogenerating the metadata. There is also some minor logic in the Aggregator to make sure the many events are grouped all pretty-like.


lusbenjamin:
#### Before
image

#### After
image

PR #892, opened 2020-02-11
Misc ListItemDeck bugfixes

lusbenjamin:
- removes deprecated `ListItem` code
- restores the ability to add new custom deck tags
- restores the italic style to the format tag bubble
- restores the ability to remove custom deck tags
- only shows the archive button on custom decks
- bugfix missing React `key` prop in `WildcardsCost`
- bugfix archive button on `ListItemDeck` now works


src/window_main/components/WildcardsCost.tsx
```diff
@@ -51,12 +55,10 @@ export default function WildcardsCost(props: WildcardsCostProps): JSX.Element {
           }
         }
       )}
-      {drawCost ? (
+      {drawCost && (
```
Manuel-777: I always forget about `&&` .__.

PR #896, opened 2020-02-12
add boosters filter preset to collection page

lusbenjamin:
### Motivation
Adds a super-useful and often requested single-click solution to the Collection page.

### DIFF
--- a/src/window_main/components/collection/CollectionTableControls.tsx
+++ b/src/window_main/components/collection/CollectionTableControls.tsx
@@ -14,6 +14,9 @@ import PagingControls from "../tables/PagingControls";
 import { defaultRarity } from "./filters";
 import { CollectionTableControlsProps } from "./types";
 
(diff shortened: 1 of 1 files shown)


src/window_main/components/collection/CollectionTableControls.tsx
```diff
@@ -14,6 +14,9 @@ import PagingControls from "../tables/PagingControls";
 import { defaultRarity } from "./filters";
 import { CollectionTableControlsProps } from "./types";
 
+const boostersFilters = (): FilterValue[] => [
+  { id: "booster", value: { true: true, false: false } }
```
Thaoden: This looks funny ;) What does `value` expect?

Manuel-777: Indeed it does.. I'm merging this but now I'm curious about the filters system..

PR #898, opened 2020-02-16
React Collection completion stats

lusbenjamin:
### Motivation
Upgrades the Collection page Sets View completion stats to React functional components and hooks, reformats the UX to be clearer at-a-glance, and adds a "Future Boosters" input.

image

#### Background
My initial build of the new "Sets View" simply wrapped the old legacy display code. Then when I added the "Draft Estimator" feature, I simply copy-pasta'd more horrible old display code. This PR cleans up that mess of `createDiv`s and replaces it with React functional components and hooks.

PS - this fixes the "refresh flicker bug" for the right-side pane inputs that @Thaoden found during initial testing


src/window_main/components/collection/CollectionStatsPanel.tsx
```diff
@@ -0,0 +1,290 @@
...
+              >
+                Completion* <CalendarSymbol />:
+              </div>
+              <label
+                className={"but_container_label"}
+                title={"rare picks per draft"}
+              >
+                <RaritySymbol rarity={"rare"} /> rares/draft:
+                <div className={"input_container"} style={inputStyle}>
+                  <input
```
Manuel-777: Could this benefit from the standarized fields used in React Settings?

src/window_main/components/collection/CollectionStatsPanel.tsx
```diff
@@ -0,0 +1,290 @@
...
+              <div className={"settings_note"} style={{ opacity: "0.6" }}>
+                <i>
+                  <a
+                    onClick={(): Promise<void> =>
+                      shell.openExternal(
+                        "https://www.mtggoldfish.com/articles/collecting-mtg-arena-part-1-of-2"
+                      )
+                    }
+                  >
+                    *[original by caliban on mtggoldfish]
```
Manuel-777: It does not look like a link, its missing the style

src/window_main/components/collection/CollectionTable.tsx
```diff
@@ -90,22 +68,30 @@ function renderHeatMaps(container: HTMLElement, stats: CollectionStats): void {
...
-    case COLLECTION_CHART_MODE:
-      renderHeatMaps(container, stats);
-      break;
-    case COLLECTION_SETS_MODE:
-      renderSetView(container, stats, setClickCallback);
-      break;
+  renderHeatMaps(container, stats);
+}
+
+function isBoosterMathValid(filters: Filters<CardsData>): boolean {
```
Manuel-777: We use this math to check weter to draw or not the collection completion stats , but in reality we should check if the set has boosters available instead.

src/window_main/components/collection/CollectionStatsPanel.tsx
lusbenjamin: Yes! I would have attempted to use the `Input` component if it were merged already. Similarly, the React Settings `Input` component could potentially use the `hooks.useBlurOnEnter` hook written in this PR.

src/window_main/components/collection/CollectionTable.tsx
lusbenjamin: We have several different potential layers of filters on the Collection page, but I am pretty confident that I have them correctly sorted:
1. The `CollectionTab` does an initial filter for `card.collectible`. This always affects all Collection views.
2. The "Boosters" preset can optionally add an additional filter for "has boosters available". This should only be applied when the user wants to use this filter; some users need to use the Collection page to find non-booster craftable cards. Additional booster math only makes sense on this domain and is only visible when the table matches the correct filter settings.
3. The "Sets View" list of sets does one additional layer of filtering, only showing Sets that "have boosters" available. (This should be fine for now, since there aren't any available "crafting only" cards by themselves in a set.)

Manuel-777: Showing collection stats for sets, even if the filtering is correct, seems wrong visually.. Currently lots of pointless sets with empty stats are shown if no filters are applied (my default view)
image

lusbenjamin: I think I understand. The problem is not the existing 3 layers of filters, it's that we need a 4th. The completion stats component should only be displayed for sets that have boosters, regardless of the current table filters being valid. I can definitely add this!

lusbenjamin: image

PR #899, opened 2020-02-17
Add negative handling for Vault

SabreCat:
Previously, the Vault progress indicator in the Economy screen would show the upward delta arrow even if the user cashed in the Vault causing the percentage progress to be negative:

image

Now, the Vault economy record includes `deltaDownContent` instead of `deltaUpContent` when appropriate, like the other headings.


lusbenjamin:
Good catch on the vault progress regression! Note that this PR takes a slightly different direction from the previous behavior:
 - Old version: present only positive vault progress (infer deltas >100 using modulo arithmetic)
 - this PR: present direct positive or negative vault progress

PR #906, opened 2020-02-23
React home tab

Manuel-777:
Converts Home to React, also cleans it up a little bit;

- Removes Discord login option and indicator.
- Removes Tournament stuff.
- Does not clean up background for discord or tournament settings/data.

Edit: This PR was stacked on #894


lusbenjamin:
@Manuel-777 heads up, plz pull before doing anything! I moved your commit to `develop` to avoid some bizarre merge cruft that was on this PR before.

src/window_main/HomeTab.tsx
```diff
@@ -0,0 +1,249 @@
...
+  wildcards: WildcardsChange[] | undefined;
+  filteredSet: string;
+  usersActive: number;
+}
+
+function HomeTab(props: HomeTabProps): JSX.Element {
+  const [filteredSet, setFilteredSet] = React.useState(props.filteredSet);
+  const orderedSets: string[] = Object.keys(db.sets).filter(
+    set => db.sets[set].collation > 0
+  );
```
lusbenjamin:
```suggestion
  const orderedSets: string[] = db.sortedSetCodes.filter(
    set => db.sets[set].collation > 0
  );
```

src/window_main/HomeTab.tsx
```diff
@@ -0,0 +1,249 @@
...
+    ipcSend("request_home", set);
+  };
+
+  orderedSets.sort((a: string, b: string) => {
+    const aSet = db.sets[a];
+    const bSet = db.sets[b];
+    if (aSet.release < bSet.release) return 1;
+    if (aSet.release > bSet.release) return -1;
+    return 0;
+  });
```
lusbenjamin:
```suggestion
```
^ can remove this if we use `db.sortedSetCodes`

src/window_main/HomeTab.tsx
```diff
@@ -0,0 +1,249 @@
...
+      <div className={lineDark} style={{ gridArea: `1 / 5 / auto / 6` }}>
+        Amount
+      </div>
+      <div className={lineDark} style={{ gridArea: `1 / 6 / auto / 8` }}></div>
+      {wildcards.map((wc: WildcardsChange, index: number) => {
+        const card = db.card(wc.grpId);
+        const ld = index % 2 ? "line_dark" : "line_light";
+
+        return card ? (
+          <>
```
lusbenjamin: This `JSX.Element` should NOT be a fragment because you are generating it inside a `map` call. This currently causes React to create a bunch of keyless nodes and causes performance issues. The fix is to replace the fragment with a `<span key={index}>...</span>` or extract the fragment into a new `WildcardsRow` component that can have a key.

Manuel-777: I'd go with the later! I certainly didn't know that, I always thought fragments were ignored :)

Manuel-777: Reading a little further, wouldn't a key on the fragment fix this?

lusbenjamin: LOL, probably! TIL

lusbenjamin: As long as you can get the console error message to go away, you are "doing it correctly"

PR #910, opened 2020-02-27
Convert shared util to TypeScript

lusbenjamin:
This work was originally part of #821. This PR extracts just the `shared/util` conversion and the bugfixes related to proper types. (I have tested all of the affected areas and confirmed that `InternalDeck` is correct for now, even if we eventually plan on using `Deck` everywhere.)


Manuel-777:
Is `Ulti.tsx` the exact same as seen in #821 ? I want to make sure this one doesn't conflict much with `Full-app-context`, particularly the bigfixes since I began using `Deck` there.

lusbenjamin:
Almost. `util.js` in current stable (v3.0.8) deals entirely with `InternalDeck` objects, and this PR only translates `util` to TS and correctly types them as such. On the other hand, the `util.tsx` in #821 does NOT deal entirely with `InternalDeck` objects, and introduced a few bugs where one part of the code was expecting an `InternalDeck` but was being handed a `Deck` or vice-versa.

I do not know if `Full-app-context` inherited any of the misleading Deck-related bugs that I still think are in #821, but I would strongly favor the version of `util.tsx` in this PR (it works with current stable) over the version of util in #821 (it doesn't work and introduces garbage data into the local DB).

PR #912, opened 2020-03-02
Store and save logged decklist order separately

dougmill:


src/shared/PlayerData.ts
```diff
@@ -78,7 +78,7 @@ const overlayCfg = {
   type_counts: false
 };
 
-const primaryBounds = remote.screen.getPrimaryDisplay().bounds;
+const primaryBounds = remote ? remote.screen.getPrimaryDisplay().bounds : {};
```
lusbenjamin: TS doesn't like this line now. I'm guessing it's just here for the unit test environment? Can we get away with some simple defaults?
```suggestion
const primaryBounds: Electron.Rectangle = remote
  ? remote.screen.getPrimaryDisplay().bounds
  : { width: 800, height: 600, x: 0, y: 0 };
```

src/shared/cardsList.ts
```diff
@@ -29,51 +30,79 @@ interface ColorsCount {
 }
 
 class CardsList {
+  private arenaLoggedList: v2cardsList;
```
Manuel-777: I think rather than including this in `CardsList` it would make more sense to create an additional set of lists in `Deck`? or separate the logic somehow? (extending the class sounds logical)

src/shared/deck.ts
```diff
@@ -28,13 +28,22 @@ class Deck {
 
   constructor(
     mtgaDeck: Partial<InternalDeck> = {},
-    main: anyCardsList = [],
-    side: anyCardsList = []
+    main?: anyCardsList,
+    side?: anyCardsList,
+    arenaMain?: anyCardsList,
```
Manuel-777: Not sure if you need it to be `anyCardsList`, I think the logs are all ``v3CardsList` now.

dougmill: `Deck.clone()` passes `v2cardsList`.

src/shared/cardsList.ts
dougmill: Moved to `Deck`.

src/types/Deck.ts
```diff
@@ -57,8 +59,8 @@ export type v2cardsList = Array<CardObject>;
 export type v3cardsList = Array<number>;
 
 export function isV2CardsList(
-  list: v2cardsList | v3cardsList
-): list is v2cardsList {
+  list: Readonly<anyCardsList>
+): list is Readonly<v2cardsList> {
```
dougmill: I checked, this can still be called with non-`Readonly` objects, and has the same effect on type analysis for them as before.

PR #916, opened 2020-03-07
Full React

Manuel-777:
_(Formerly "Full App Context")_

_In the beginning, there was just html._

I went ahead a few steps and converted all remaining components to React, mainly because they required some form of context to work. I initially used the new Context API from React, but quickly moved into Redux because speed.
Some goodies that come with this power:
- Card Hover hook.
- Other hooks like Color Picker are now full react and can be easily deployed.
- Some of the new components from the website (like the draft viewer) were easily imported and adapted.
- A lot of deprecated code was removed, still more to go.
- Lots of `PlayerData` was and can be moved into Redux.
- New components. Lots of them!

BUT, there is still a lot of work to do! (taken from my ~~paper sheet~~ to-do list):
- [x] Show version number on main.
- [x] Card Hover ownership stars.
- [x] Missing wildcards count in `DeckList`
- [x] React missing log popup.
- [x] React share popups.
- [x] React keyboard shortcuts popup.
- [x] Background color / visual settings change.
- [x] Convert stats-panel.
- [x] Convert collection completion heatmap.

And some bugs:
- [x] Matches history big  memory leak.
- [x] Ranks in Explore are not sorted.
- [x] There is a race condition when you scroll Explore apparently.
- [x] Navigate draft replays with arrow keys.
- [x] Economy rows horizontal overflow.
- [x] Settings selectors don't close when clicked.
- [x] Arena log input on settings does not update
- [x] Ranked icons do not navigate correctly
- [x] Deleting/adding deck tags is broken (requires manual "refresh")
- [x] Deleting/adding match tags is broken (requires manual "refresh")
- [x] Fancy new settings toggles do not work

Stuff to tackle in future PRs:
- massive folder re-org / standardize file and class names
- deprecate `PlayerData` / make main window respond to Arena without "manual refresh"
- Restore `MatchResultsStatsPanel` to `DeckView`
- consolidate duplicated Components in `main` and `overlay`
- clean up deprecated styles?

I am sure there are improvements to make, like using more `useCallback` when needed (and not when not needed), and maybe use more `useMemo`'s. For instance, I made hover a lot smoother by removing the function from the component and making it pass the `grpId` as an argument instead of using the callback hook.

Also, things like popups can be rewritten and redesigned entirely, things **don't have** to stay the same as before, especially if there is a cleaner way ;)


src/window_background/globals.ts
```diff
@@ -18,7 +18,7 @@ let currentDeck = new Deck();
 
 const debugLog = false;
 
-const debugNet = false;
+const debugNet = true;
```
lusbenjamin: TODO: set this back to `false` before merging

src/window_main/app/ipcListeners.tsx
```diff
@@ -0,0 +1,175 @@
...
+    dispatchAction(dispatcher, SET_OFFLINE, true);
+  });
+
+  ipc.on("show_loading", (): void => {
+    dispatchAction(dispatcher, SET_LOADING, true);
+  });
+
+  ipc.on("hide_loading", (): void => {
+    dispatchAction(dispatcher, SET_LOADING, false);
+  });
```
lusbenjamin: these are currently unused

src/window_background/httpApi.ts
```diff
@@ -143,7 +144,7 @@ function handleNotificationsResponse(
   // Like, check if arena is open at all, if we are in a tourney, if we
   // just submitted some data that requires notification pull, etc
   // Based on that adjust the timeout for the next pull.
-  setTimeout(httpNotificationsPull, 10000);
+  //setTimeout(httpNotificationsPull, 10000);
```
lusbenjamin: is this deliberate or just a temp change for dev?

src/window_background/httpApi.ts
```diff
@@ -255,9 +257,8 @@ function handleAuthResponse(
     } else {
       ipcLog("No need to fetch remote player items.");
     }
-    httpNotificationsPull();
+    //httpNotificationsPull();
```
lusbenjamin: is this deliberate or just a temp change for dev?

src/main.js
```diff
@@ -64,6 +62,7 @@ app.on("ready", () => {
   if (app.isPackaged) {
     startUpdater();
   } else {
+    installDevTools();
```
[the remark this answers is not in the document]
Manuel-777: I thought about this, I think it's fine if end users get devtools it also not what everyone expects (it's open source, people can download source and compile if they're up to it)
I am totally okay moving it to dev only, this was because the new devtools installer is imported always.

src/shared/ReactSelect.tsx
```diff
@@ -43,20 +43,21 @@ export function ReactSelect(props: ReactSelectProps): JSX.Element {
...
-                  onClick={onClickOption}
-                >
-                  {formatterFunc(option)}
-                </button>
-              );
-            })}
+          {props.options.map(option => {
+            return typeof option == "string" && option.startsWith("%%") ? (
+              <div className="select_title">{option.replace("%%", "")}</div>
+            ) : (
```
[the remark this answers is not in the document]
Manuel-777: Yes, this is a trick to display titles inside the select options, adding "%%" to one of the options. It's used in explore to make a division between active and archived event names.
The proper way would require using an array of objects as the options instead, and have each option be a separator or text, etc.
Also had to add a typeof clause because sometimes we pass objects as options and format them with the formatter function.

src/overlay/reducers.tsx
```diff
@@ -0,0 +1,40 @@
+import { combineReducers } from "redux";
+import {
+  Action,
+  SET_HOVER_IN,
+  SET_HOVER_OUT,
+  SET_HOVER_SIZE,
+  HoverState
+} from "../window_main/app/reducers";
```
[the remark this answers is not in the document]
Manuel-777: Yes, definitely!

src/window_background/httpApi.ts
Manuel-777: Deliberate, I'm not these will be used anymore. They were only useful for discord auth and that was removed.

Manuel-777:
> Review Part 1 (shared, overlay, background code):
> 
> * THIS IS SO AWESOME!!!
> * how close are we to removing the legacy `dom-fns` like `createDiv`?
> * can we finally merge `ReactSelect` and `WrappedReactSelect`?
> 
> (more comments and feedback soon in Review Part 2)

Forgot to reply these;
We can deprecate `WrappedReactSelect` already? I think that and the legacy Dom functions are used in the sidebar only, which is mounted into React.

src/window_main/components/ShareButton.tsx
```diff
@@ -12,21 +20,33 @@ export default function ShareButton({
...
-      createShareDialog(shareExpire => deckShareLink(data, shareExpire));
+      const deckString = JSON.stringify(data);
+      dispatchAction(dispatcher, SET_SHARE_DIALOG, {
+        data: deckString,
+        type
+      });
     } else if (type == "actionlog") {
-      createShareDialog(shareExpire => logShareLink(data, shareExpire));
+      dispatchAction(dispatcher, SET_SHARE_DIALOG, {
+        data: data,
```
lusbenjamin: Noticed that the "actionLog" `else if` branch is the only one that does not call `JSON.stringify(data)` on the data first. Is this deliberate?

src/window_main/components/popups/OutputLogInput.tsx
```diff
@@ -0,0 +1,90 @@
...
+    });
+    if (paths && paths.length && paths[0]) {
+      setLog(paths[0]);
+    }
+  }, []);
+
+  useEffect(() => {
+    // React doesnt give css time to know there was a change
+    // in the properties, adding a timeout solves that.
+    setTimeout(() => {
```
lusbenjamin: would `useLayoutEffect` handle these cases more correctly? (Never used it myself, but I remembered [reading about it here](https://reactjs.org/docs/hooks-reference.html#uselayouteffect)

src/window_main/components/popups/Share.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const expireOptions = ["One day", "One week", "One month", "Never"];
+
+  useEffect(() => {
+    setUrl(url);
+  }, [url]);
+
+  useEffect(() => {
+    // React doesnt give css time to know there was a change
+    // in the properties, adding a timeout solves that.
+    setTimeout(() => {
```
lusbenjamin: ditto

src/window_main/components/popups/EditKey.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const reportStr = keys.join("+");
+    setKeyDesc(reportStr);
+    zEvent.stopPropagation();
+    zEvent.preventDefault();
+  }, []);
+
+  useEffect(() => {
+    // React doesnt give css time to know there was a change
+    // in the properties, adding a timeout solves that.
+    setTimeout(() => {
```
lusbenjamin: mebbe `useLayoutEffect`?

src/window_main/components/matches/types.ts
```diff
@@ -44,7 +44,7 @@ export interface MatchesTableProps {
   editTagCallback: (tag: string, color: string) => void;
   events: string[];
   filterDataCallback: (data: MatchTableData[]) => void;
-  openMatchCallback: (matchId: string | number) => void;
+  openMatchCallback: (match: any) => void;
```
lusbenjamin:
```suggestion
  openMatchCallback: (match: InternalMatch) => void;
```

src/window_main/components/matches/types.ts
```diff
@@ -61,7 +61,7 @@ export interface MatchesTableControlsProps
 export interface ListItemMatchProps {
   tags: TagCounts;
   match: InternalMatch;
-  openMatchCallback: (matchId: string | number) => void;
+  openMatchCallback: (match: any) => void;
```
lusbenjamin:
```suggestion
  openMatchCallback: (match: InternalMatch) => void;
```

src/window_main/components/decks/types.ts
```diff
@@ -49,7 +49,7 @@ export interface DecksTableProps {
   deleteTagCallback: (deckid: string, tag: string) => void;
   editTagCallback: (tag: string, color: string) => void;
   filterDataCallback: (data: DecksData[]) => void;
-  openDeckCallback: (id: string | number) => void;
+  openDeckCallback: (deck: any) => void;
```
lusbenjamin:
```suggestion
  openDeckCallback: (deck: InternalDeck) => void;
```

src/window_main/components/decks/types.ts
```diff
@@ -63,7 +63,7 @@ export interface DecksTableControlsProps extends TableControlsProps<DecksData> {
 
 export interface DecksTableRowProps extends TableViewRowProps<DecksData> {
   tags: TagCounts;
-  openDeckCallback: (id: string | number) => void;
+  openDeckCallback: (deck: any) => void;
```
lusbenjamin:
```suggestion
  openDeckCallback: (deck: InternalDeck) => void;
```

src/window_main/HomeTab.tsx
```diff
@@ -1,25 +1,28 @@
...
+  const usersActive = useSelector(
+    (state: AppState) => state.homeData.usersActive
+  );
 
-function HomeTab(props: HomeTabProps): JSX.Element {
-  const [filteredSet, setFilteredSet] = React.useState(props.filteredSet);
-  const orderedSets: string[] = db.sortedSetCodes.filter(
+  const [filteredSet, setFilteredSet] = React.useState(fSet);
+  const [requested, setRequested] = React.useState(false);
+  const orderedSets: string[] = Object.keys(db.sets).filter(
```
lusbenjamin: possible merge regression

src/window_main/HomeTab.tsx
```diff
@@ -49,11 +52,27 @@ function HomeTab(props: HomeTabProps): JSX.Element {
...
+    }
+  }, [requested, usersActive, filteredSet]);
+
+  orderedSets.sort((a: string, b: string) => {
+    const aSet = db.sets[a];
+    const bSet = db.sets[b];
+    if (aSet.release < bSet.release) return 1;
+    if (aSet.release > bSet.release) return -1;
+    return 0;
+  });
```
lusbenjamin: possible merge regression

src/window_main/Auth.tsx
```diff
@@ -0,0 +1,150 @@
...
+          <label className="form-label">Email</label>
+          <div className="form-input-container">
+            <input
+              onChange={handleEmailChange}
+              type="email"
+              id="signin_email"
+              autoComplete="off"
+              value={authForm.email}
+            />
+          </div>
```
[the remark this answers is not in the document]
Manuel-777: Maybe, I would have to figure a way to style them all consistently, currently form inputs have a very different style from the rest of the app.

src/window_main/app/appState.tsx
```diff
@@ -0,0 +1,120 @@
...
+    filterSort: string;
+    filterSortDir: -1 | 1;
+    filteredMana: number[];
+    filteredRanks: string[];
+    filterSkip: number;
+  };
+}
+
+export const defaultState: AppState = {
+  topArtist: "Bedevil by Seb Seb McKinnon",
```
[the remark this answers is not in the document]
Manuel-777: OMG

src/window_main/components/ShareButton.tsx
Manuel-777: Its preconverted to base64, so its ready to share. But that did uncovered a little bug specific to this button :D

src/window_main/components/popups/OutputLogInput.tsx
Manuel-777: By the description it would seem so, but I tried and it didn't work. I believe its because of this;
> Updates scheduled inside useLayoutEffect will be flushed synchronously, **before** the browser has a chance to paint.

I guess we need to let the browser paint at least once before making the CSS diff-trick to animate.

PR #919, opened 2020-03-12
bugfix overlay card hover logic

lusbenjamin:
This PR fixes the card hover feature in draft and deck lists in the overlays by fully converting the remaining pre-Redux card hover logic into Redux + `useHoverCard`. It also converts the `ActionLog` and cleans up the callback cruft.


lusbenjamin:
I saw the one in `window_main`, but was mostly focused on fixing the overlays instead of consolidating or moving components. Feel free to clean up the folders now!

PR #920, opened 2020-03-13
Tables fixes

Manuel-777:
I will be pushing all fixes I see necessary to the tables logic here as I go.


src/window_main/components/collection/CollectionTable.tsx
```diff
@@ -74,27 +69,31 @@ function isBoosterMathValid(filters: Filters<CardsData>): boolean {
...
+  openCardCallback
 }: CollectionTableProps): JSX.Element {
   const [tableMode, setTableMode] = React.useState(cachedTableMode);
   const [countMode, setCountMode] = React.useState(ALL_CARDS);
   const [rareDraftFactor, setRareDraftFactor] = React.useState(3);
   const [mythicDraftFactor, setMythicDraftFactor] = React.useState(0.14);
   const [boosterWinFactor, setBoosterWinFactor] = React.useState(1.2);
   const [futureBoosters, setFutureBoosters] = React.useState(0);
-  const data = useMemo(() => getCollectionData(), []);
-  const sortedSetCodes = useMemo(() => db.sortedSetCodes, []);
```
Manuel-777: Why removing this? it caused a loop when using the set filter sort

PR #934, opened 2020-03-17
Draft view polish

Manuel-777:
Some CSS and Div alignments on the draft replay view, also jammed in a little something for sliders;

image

### DIFF
--- a/src/window_main/components/misc/Slider.tsx
+++ b/src/window_main/components/misc/Slider.tsx
@@ -1,22 +1,16 @@
 /* eslint-disable react/prop-types */
 import React, { useState } from "react";
 
-interface SliderPosition {
-  text: string;
-  hide: boolean;
-  color: string;
-}
+export class SliderPosition {
+  public text: string;
+  public hide: boolean;
+  public color: string;
 
-export function sliderPosition(
-  text = "",
-  hide = false,
-  color = "var(--color-light-50)"
-): SliderPosition {
-  return {
-    text: text,
-    hide: hide,
-    color: color
-  };
+  constructor(_text = "", _hide = false, _color = "var(--color-light-50)") {
+    this.text = _text;
+    this.hide = _hide;
+    this.color = _color;
+  }
 }
 
 interface SliderProps {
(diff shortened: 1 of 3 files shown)


src/window_main/components/misc/Slider.tsx
```diff
@@ -33,14 +57,12 @@ export default function Slider(props: SliderProps): JSX.Element {
...
 
-  const stepsNumber = (max - min) / step;
-
   React.useEffect(() => {
     setValue(props.value);
   }, [props.value]);
 
   return (
-    <div className="slidecontainer">
+    <div style={{ ...props.containerStyle }} className="slidecontainer">
```
lusbenjamin: Nit: this should probably just `style={props.containerStyle}` to prevent unnecessary rerenders for the same style prop (has no practical performance implications in this case)

src/window_main/components/misc/Slider.tsx
```diff
@@ -52,13 +74,21 @@ export default function Slider(props: SliderProps): JSX.Element {
...
+            <div className="slider_mark_outer" key={c.text + "-" + i}>
+              <div
+                className="slider_mark_hor"
+                style={{ backgroundColor: c.color, opacity: c.hide ? 0 : 1 }}
+              />
+              {c.text !== "" ? (
+                <div className="slider_mark_text">{c.text}</div>
+              ) : (
+                <></>
+              )}
```
lusbenjamin: Nit: remove "empty" final clause in ternary by just using boolean `&&`:
```suggestion
              {c.text !== "" && (
                <div className="slider_mark_text">{c.text}</div>
              )}
```

src/window_main/components/misc/Slider.tsx
Manuel-777: I wanted to allow an undefined style without adding an if here, but I agree it's probably the same.

Manuel-777:
> Lookin' so Fancy wow!
> 
> left a few totally minor comments. Also, curious why you chose the interface+function combo for `SliderPosition` instead of just using a single `class`?

I haven't thought about a class, I guess it makes more sense that way!

PR #935, opened 2020-03-18
Redux Toolkit

lusbenjamin:
### Motivation
We are about to convert all of the remaining application state into Redux (I hope). Adding [Redux Toolkit](https://redux-toolkit.js.org/) to our stack before we do this is probably a very good idea in terms of reducing potential mistakes and using best practices from the start.

This PR:
- adds the `redux-toolkit` package dependency
- sets up `configureStore` in the display processes
- converts `topNav` into a slice to demonstrate `createSlice`
- refactors the Explore page state into a proper slice
  - bugfix: `activeEvents` should be an Array
  - upgrade: better logic for determining when a query has no more results

Highly recommended reading:
- https://redux-toolkit.js.org/tutorials/basic-tutorial
- https://redux-toolkit.js.org/usage/usage-with-typescript

### DIFF
--- a/src/shared/redux/reducers.tsx
+++ b/src/shared/redux/reducers.tsx
@@ -13,7 +13,6 @@ export const SET_HOVER_IN = "SET_HOVER_IN";
 export const SET_HOVER_OUT = "SET_HOVER_OUT";
 export const SET_HOVER_SIZE = "SET_HOVER_SIZE";
 export const SET_OFFLINE = "SET_OFFLINE";
-export const SET_LOADING = "SET_LOADING";
 export const SET_SUB_NAV = "SET_SUB_NAV";
 export const SET_LOGIN_STATE = "SET_LOGIN_STATE";
 export const SET_LOGIN_FORM = "SET_LOGIN_FORM";
@@ -130,17 +129,13 @@ const offline = (
   }
 };
 
-const loading = (
-  state: boolean = defaultState.loading,
-  action: Action
-): boolean => {
-  switch (action.type) {
-    case SET_LOADING:
-      return action.value;
-    default:
-      return state;
+export const loadingSlice = createSlice({
+  name: "loading",
+  initialState: defaultState.loading,
+  reducers: {
+    setLoading: (state, action): boolean => action.payload
   }
-};
+})
 
 export const topNavSlice = createSlice({
   name: "topNav",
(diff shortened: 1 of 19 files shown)


Manuel-777:
I'm digging this, One of the reasons I made the `dispatchAction` function was this, and wasnt sure if it was correct at all to "interfere" in the default flow of Redux like that :D

lusbenjamin:
As 1 of the 3 major 3rd-party packages (Redux, React-Redux, Redux Toolkit), these conventions are highly likely to futureproof / understood by new contributors familiar w Redux. (Also Redux Toolkit is part of Create React App.)

src/window_main/app/ipcListeners.tsx
```diff
@@ -165,11 +166,7 @@ export default function ipcListeners(dispatcher: unknown): void {
   });
 
   ipc.on("settings_updated", (): void => {
-    dispatchAction(
-      dispatcher,
-      SET_TOP_NAV,
-      pd.settings.last_open_tab || MAIN_HOME
-    );
+    dispatcher(setTopNav(pd.settings.last_open_tab ?? MAIN_HOME));
```
Manuel-777: Maybe I just dont understand the null coalescing properly, but shouldn't `pd.settings.last_open_tab` be a null for it to return `MAIN_HOME` as a default? currently it cant be other than a number or undefined.

lusbenjamin: coalescing handles both `null` and `undefined`: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

src/shared/redux/reducers.tsx
```diff
@@ -351,18 +351,6 @@ const shareDialog = (
   }
 };
 
-const activeEvents = (
-  state: string[] = defaultState.activeEvents,
-  action: Action
-): string[] => {
-  switch (action.type) {
-    case SET_ACTIVE_EVENTS:
-      return { ...state, ...action.value };
```
lusbenjamin: this was a bug; the intended return value was probably `return [ ...state, ...action.value ];`

Manuel-777: Ohhh, great catch! I think this was the source of the array conversion issues I was having !

Manuel-777:
My attempt at learning this was sucessful! Should we go and convert the whole reducer here or in separate PRs?

PR #940, opened 2020-03-20
Extract settings slice

Manuel-777:
An attempt to improve overlays response time and smoothness, mostly.

Seems to do the trick, but I still see some odd behaviour on the overlays (like, updating some settings seems to do nothing for a while when you start the app)


Manuel-777:
deprecated in favor of #941

PR #944, opened 2020-03-23
bugfix NeDB ensure compaction run at launch

lusbenjamin:
### Motivation
We had a report of a user with a personal DB file over 1GB that caused the tool to be unable to boot. The file included lots and lots of duplication, and could be fixed by "manually" compacting it. We do not know how the DB file got to be so big without auto compact fixing it, and I was unable to cause similar run-away-file-size behavior when I attempted to force it (rapidly starting/stopping the tool without giving compaction time to run).

Although this is still basically a shot in the dark, this PR forces the DB the run compaction as the first thing it does after knowing where the DB file is. These changes are designed to add an additional safety to prevent similar issues from happening if possible.

### DIFF
--- a/src/shared/db/NeDbDatabase.ts
+++ b/src/shared/db/NeDbDatabase.ts
@@ -65,9 +65,11 @@ export class NeDbDatabase implements LocalDatabase {
   init(dbName: string, arenaName?: string) {
     this.dbName = sanitize(arenaName ? arenaName : dbName);
     this.datastore = new Datastore({
-      filename: path.join(USER_DATA_DIR, this.dbName + ".db"),
-      autoload: true
+      filename: path.join(USER_DATA_DIR, this.dbName + ".db")
     });
+    // ensure session begins with most compact possible db
+    this.datastore.persistence.compactDatafile();
+    this.datastore.loadDatabase();
     // async wrappers of datastore methods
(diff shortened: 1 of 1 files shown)


lusbenjamin:
🤔 Not sure this PR does anything at all. NeDBs docs literally say:
>  The database is automatically compacted (i.e. put back in the one-line-per-document format) every time you load each database within your application.

lusbenjamin:
Some ideas for more effective safeguards:
- force a compaction at certain intervals during log first pass
- force a compaction at the end of doing log first pass
- force a compaction at the end of doing patreon remote sync
- refactor the database into individual collection files, one per major object type (this isolates changes to limit DB churn and of course results in smaller overall files, kind of "quick and dirty clustering")

Manuel-777:
Did compaction on Nedb actually removed the dupes? because the dupes problem is definitely because we are inserting a new deck when read instead of upserting it, nedb doesnt care if there is a duplicate index so it creates a new one.

I dont know if nedb has a unique index feature we can use to prevent this?

lusbenjamin:
I don't think we need to take any action regarding upserts: I really am highly highly confident in that logic by now. I BELIEVE THE DUPLICATION IS NEDB WORKING CORRECTLY. But I do think there are probably ways we can get it do be more efficient overall.

PR #950, opened 2020-03-24
Bugfix archive buttons

lusbenjamin:
### Motivation
This innocuous sounding PR has the modest goal of fixing the archive and unarchive button behavior.

### Approach
REFACTOR ALL THE THINGS!

### DIFF
--- a/src/shared/redux/reducers.tsx
+++ b/src/shared/redux/reducers.tsx
@@ -1,9 +1,16 @@
 /* eslint-disable @typescript-eslint/camelcase */
 import { createSlice } from "@reduxjs/toolkit";
+import _ from "lodash";
+import { TableState } from "react-table";
 import { combineReducers } from "redux";
 import { MergedSettings } from "../../types/settings";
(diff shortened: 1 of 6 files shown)


Manuel-777:
From discord;

> I noticed the PR only updates state in the UI , not anything that comes from the background?
like, archiving is only done in the UI despite the result of the operation on the actual store/background.
> should we do that or you just did it that way until we do the full round-around redux state/ipc heavy work?
> Archiving is not a worry in that sense, because its basically an on/off toggle that has little impact. Just curious if thats something you thought about.

Also, I noticed theres a crash in History tab, not sure if its related , nor if all other tabs need to be converted now, before this lands?

lusbenjamin:
Nothing needs to "come from the background" in this PR because it goes all the way to putting the player data state into Redux and simply updating the archived value there. (It still sends an IPC update to keep background in sync).

...buuuuuut this PR is too ambitious. Closing in favor of building a simple cache for now.

PR #952, opened 2020-03-25
Timeline Tab

Manuel-777:
Something I had on the stash for a long time, it allows you to visualize your ranked progress during this season in a graph, much like other trackers do;

image

It still has some lacking features or things that need to be tested:
- [ ] I dont play that much to reach Mythic and make the graph grow properly after that (percentile and mythic rank), With a `.db` from a user that reached it would be enough
- [x] Match details are a bit lacking, maybe a button to view the match itself?
- [ ] Show winrates of the hovered deck's period?
- [x] Allow to view previous Seasons, maybe trough a select or an arrow
- [ ] Get the best most effective deck's version?
- [ ] Make the Rank icons on top open this screen rather than History?
- [ ] Show most prevalent color combinations/archetypes for each rank tier, regardless of our deck.

Note I do not plan to remove ranked history from history tab, that should stay as it is.


Manuel-777:
Merged to test and because it now includes some important changes to the Seasonal dataset.

PR #958, opened 2020-03-25
force auto refresh w playerDataTimestamp

lusbenjamin:


Manuel-777:
I like this!, its stylish in its hackiness!

PR #959, opened 2020-03-25
convert background to ts

lusbenjamin:


src/window_background/background.ts
```diff
@@ -53,9 +56,9 @@ if (!fs.existsSync(globals.actionLogDir)) {
 globals.toolVersion = (app || remote.app)
   .getVersion()
   .split(".")
-  .reduce((acc, cur) => +acc * 256 + +cur);
+  .reduce((acc, cur) => +acc * 256 + +cur + "");
```
Manuel-777: `toolVersion` should be a number, not a string, so maybe `parseInt(acc)` ? (also change its type definition)

PR #961, opened 2020-03-26
Async GRE parser

Manuel-777:
This should have no impact in the parsing itself, but it would allow the GRE parser to run on renderer without blocking the process. (I havent really tested that part yet)

### DIFF
--- a/src/window_background/greToClientInterpreter.ts
+++ b/src/window_background/greToClientInterpreter.ts
@@ -682,7 +682,10 @@ export function GREMessageByID(msgId: number, time: Date): void {
   );
 }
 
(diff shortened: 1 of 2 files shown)


lusbenjamin:
What is the longer term goal of this PR? I think it is a pre-req for some crazy plans to add Match replays to the renderer window process?

Manuel-777:
> What is the longer term goal of this PR? I think it is a pre-req for some crazy plans to add Match replays to the renderer window process?

Thats correct! I found in my crazy lab tests that the renderer being blocked while processing match data makes it unusable, even for testing.

PR #962, opened 2020-03-26
App-wide redux store test

Manuel-777:
The idea is quite simple, instead of cloning the whole state tree and calculating the diffs, just relay the actions from one process to the others that need it, using simple functions that share basic constants across all processes.

When one process wants to dispatch an action it does on its own store and relays the same action to the main store (ìn `main.js`), which at the same time relays the action to other processes, if indicated;

fancy infographic
* Black = IPC, Green = dispatcher

Notice the slices have their own folder, and then each process needs its own separate store which could be just one slice, But main process should have all the slices, always.

One thing I have considered is adding a "master" action to synchronize the entire store or big slices of data to a specific process, but I am not sure how to implement that. It could be useful in situations where we expect to process a lof of data quickly and just dispatch it at once with a single IPC message instead of a bunch of actions that might choke the ipc channels (thinking of initial loading of player data for instance).


Manuel-777:
I believe I found a way to make this work, basically when updating huge objects with lots of sub-objects and arrays into a redux store you get poor performance, this is where we use inmutability to optimize. But adding inmutability here would cause a ton of other issues.

So, my solution was to create a global store on both renderer and background that is populated trough the actions we send to redux, this way we use a single method to send the data (with an action) and there the reducer would update the global object and redux. So, we keep the "heavy" data in the store and the accessors (indexes) in redux.

PR #967, opened 2020-03-27
Economy row operators

Manuel-777:
I found some issues with the operators in EconomyRow, we were using lots of double negation and double and like ` {!!delta.wcCommonDelta && ()}` to draw components, which resulted in the results of the operations ("0") to be drawn on the economy row, or simply fail.

Not sure when this was introduced trough, economy was working fine recently.

### DIFF
--- a/src/window_background/greToClientInterpreter.ts
+++ b/src/window_background/greToClientInterpreter.ts
@@ -661,7 +661,7 @@ export function processAll(): void {
       }
     }
   }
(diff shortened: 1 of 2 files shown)


Manuel-777:
I know using `{operation ? <Component> : <></>}` sucks, but I find it the safest way to check. I always seem to have issues like that with any other kind of operator check to draw components in JSX..

PR #969, opened 2020-03-27
no auto refresh during match

lusbenjamin:
this PR attempts to solve some performance problems by preventing the auto-refresh code on the renderer process unless Arena is _NOT_ currently in a match.


Manuel-777:
Im reverting the auto refrresh changes so this one can be closed now. I think we need to do something harder than "update everything at once", but im still not sure what.

Issue #1022, opened 2020-04-22
Old Events recorded as Jan 19 1970


Thaoden:
Event timestamps before March 17 2020 are erroneously marked as Jan 19 1970. The timestamps are all Jan 19, but the events also have a time of day, ranging from 1:18 AM to 10:04 AM.

![](https://cdn.discordapp.com/attachments/467737642306371584/702618161043734528/unknown.png)

In this screenshot, events are sorted according to their dates.

![](https://cdn.discordapp.com/attachments/467737642306371584/702620292392157306/unknown.png)

This is tool version `4.0.17`.

PR #1024, opened 2020-04-23
isEpochTimestamp fix

Manuel-777:
I am not sure how non patreons could have ended with epoch timestamps unless we were forcefully downgrading Js times into regular Epoch (I can see myself doing that in the past, so why not)

Anyhow, this should fix the display issues in #1022 and include a new handy method that should probably be used just in case for a while in the rest of the datasets.


src/shared/util.ts
```diff
@@ -559,3 +559,14 @@ export function prettierDeckData(deckData: InternalDeck): InternalDeck {
...
+// Im not sure if this would be better as a converter instead of a check
+export function isEpochTimestamp(timestamp: number): boolean {
+  const asDate = new Date(timestamp);
+  // Even if we have a date in epoch between 2000 and
+  // 2100 it will be parsed as 1970.
+  if (asDate && asDate.getUTCFullYear() < 2000) {
+    return true;
+  }
+  return false;
+}
```
Thaoden, line 572:
```suggestion
return asDate.getUTCFullYear() < 2000;
```

src/window_background/httpApi.ts
```diff
@@ -84,6 +84,7 @@ function syncUserData(data: any): void {
       const id = doc._id;
       doc.id = id;
       delete doc._id;
+      if (isEpochTimestamp(doc.date)) doc.date = doc.date * 1000;
```
Thaoden, line 87:
```suggestion
if (isEpochTimestamp(doc.date)) doc.date *= 1000;
```

src/window_background/loadPlayerConfig.ts
```diff
@@ -134,6 +135,9 @@ export async function loadPlayerConfig(): Promise<void> {
     const eventsList: InternalEvent[] = savedData.courses_index
       .filter((id: string) => savedData[id])
       .map((id: string) => {
+        if (isEpochTimestamp(savedData[id].date)) {
+          savedData[id].date = savedData[id].date * 1000;
+        }
```
Thaoden, line 140:
```suggestion
        if (isEpochTimestamp(savedData[id].date)) {
          savedData[id].date *= 1000;
        }
```

I think the `*=` operator exists in TS, too?

Thaoden:
Looks good, though the issue did not happen to all non-Patreons.

Also, are the new timestamps persisted in the local user database?

Manuel-777:
> Also, are the new timestamps persisted in the local user database?

Not unless they synchronize or the event has not finished yet

src/shared/util.ts
```diff
@@ -565,8 +565,5 @@ export function isEpochTimestamp(timestamp: number): boolean {
   const asDate = new Date(timestamp);
   // Even if we have a date in epoch between 2000 and
   // 2100 it will be parsed as 1970.
-  if (asDate && asDate.getUTCFullYear() < 2000) {
-    return true;
-  }
-  return false;
+  return asDate && asDate == asDate && asDate.getUTCFullYear() < 2000;
```
Thaoden, line 568: Why the equality check?

PR #1026, opened 2020-04-24
card hover optimizations

Manuel-777:
I havent really profiled this, keeping it open to test more


Manuel-777:
Made some Profiling and this doesnt really make the whole process faster, but rather more granular, that is, instead of one big "hop" we have many smaller "hops" going on. I am still unsure while this is slow in the first place 🤷

PR #1032, opened 2020-04-24
More typings

Thaoden:
This in an attempt at removing a forced cast to `any`, that was previously necessary.

Multiple things going on here:

- `Details` only had a single optional property that was never used. Removed the type.
- Introduced some helper type (`DetailsSrcDestCategoryType`) to help with later typings
- Introduced an intersection type of all possible `DetailsXYZ` types, possibly making the union type obsolete. I've left it in for now though.
- Now to the real thing: With the intersection type in place, it is possible to switch over all possible keys in all possible `DetailsXYZ` types. In a given switch case, the compiler can infer what property of the intersection type he is looking at. In can be in any of the properties in the `KeyValuePairInfo` though, so use the helper method to extract the value from one of them and fall back to `0` if that fails.

Admittedly, it has become a lot more verbose. But it is type safe :)

I've not been able to run checks or prettier locally before committing, npm playing its tricks on me...

### DIFF
--- a/src/proto/GreTypes.ts
+++ b/src/proto/GreTypes.ts
@@ -2,7 +2,6 @@
 /* eslint-disable @typescript-eslint/no-empty-interface */
 // generated file, do not touch!
 import ByteBuffer from "bytebuffer";
-import { DetailsKeyType } from "../types/greInterpreter";
 
 export interface AIConfig {
   matchConfig?: MatchConfig;
@@ -1142,7 +1141,7 @@ export interface JoinQueueResponse {
 }
 
 export interface KeyValuePairInfo {
-  key?: DetailsKeyType;
+  key?: string;
   type?: KeyValuePairValueType;
   valueUint32: number[];
   valueInt32: number[];
--- a/src/proto/GreTypes.ts
+++ b/src/proto/GreTypes.ts
@@ -2,6 +2,7 @@
 /* eslint-disable @typescript-eslint/no-empty-interface */
 // generated file, do not touch!
 import ByteBuffer from "bytebuffer";
+import { DetailsKeyType } from "../types/greInterpreter";
 
 export interface AIConfig {
   matchConfig?: MatchConfig;
@@ -1141,7 +1142,7 @@ export interface JoinQueueResponse {
 }
 
 export interface KeyValuePairInfo {
-  key?: string;
+  key?: DetailsKeyType;
   type?: KeyValuePairValueType;
   valueUint32: number[];
   valueInt32: number[];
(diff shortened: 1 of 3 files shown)


src/proto/GreTypes.ts
```diff
@@ -1141,7 +1142,7 @@ export interface JoinQueueResponse {
 }
 
 export interface KeyValuePairInfo {
-  key?: string;
+  key?: DetailsKeyType;
```
Manuel-777, line 1145: Sorry, but this file should not be modified! Its generated from the protobuffs so each update of proto will override it.

Manuel-777:
Looking good now! The only thing I don't like about this is that adding a new property for details or a new annotation now require a *lot* more effort, since we now have more lines to modify in order to add the new type.

Also, I will investigate a littel bit more, but I think some values could end up with arrays instead of just the first value, but I supose we just have to add a function and correct types in that case?

Thaoden:
I'm fine with the switch compression.

I'm not a big fan of the extraction of all possible values in the same method, though, as they require manual casts later (I like it generally more just to let the compiler do his work). I have thus separated the different cases for `number`, `boolean` and `string` and the respective arrays in their own methods, keeping the `length === 1` logic.

And yes, adding a new type now will make every lightbulb flash red ;) But it will also tell us when we missed a case.

src/window_background/greToClientInterpreter.ts
```diff
@@ -534,31 +534,49 @@ function annotationsSwitch(ann: Annotations, type: AnnotationType): void {
...
+
+function extractBooleanArrayFromKVP(obj: KeyValuePairInfo): boolean[] {
+  return obj.valueBool;
+}
+
+function extractStringValueFromKVP(obj: KeyValuePairInfo): string | undefined {
+  return obj.valueString.length > 0 ? obj.valueString[0] : undefined;
+}
+
+function extractStringArrayFromKVP(obj: KeyValuePairInfo): string[] {
```
Manuel-777, line 576: How are these any different than the unified extractor function? we are asserting the type the moment we call the function just like we were asserting the type when we use `extractValueFromKVP() as type`.

I really really liked my unified function :D

Thaoden: Well there was a bug in the previous code when casting to `number` when the cast should have been to `number | undefined`. But telling the compiler "I know what I'm doing" (aka, casting), he doesn't complain anymore ;) Not using casts prevents these subtle bugs.
I really really _really_ like the deconstructed version that relies more hevily on the compiler. :D

Issue #419, opened 2019-06-15
Draw probability ignore known cards such as scried ones


Porkepix:
I could notice that on a game today: mulligan at the start of the game resulting then in a scry. I did put the only copy of the card I scried at the bottom of my library, so that draw probability should be 0%. But the draw chance was still the same as it was, not considering this card is known.

Manuel-777:
implemented with #1035

PR #1035, opened 2020-04-25
Remove cards on bottom

Manuel-777:
A little thing that was requested many times, scryed cards, or rather, cards that are knowingly on the bottom of the library, should be ignored from the cards odds calculation.

image

In the example I scryed to the bottom: An Island, a Temple of Mystery and a Mystical Dispute.
The only issue is that these cards are entirely removed from the cards left in the library rather than just the odds calculation (we probabaly need to add a separate store for the odds)

### DIFF
--- a/src/shared-store/currentMatchStore.ts
+++ b/src/shared-store/currentMatchStore.ts
@@ -31,6 +31,7 @@ export const matchStateObject = {
   currentDeck: new Deck(),
   originalDeck: new Deck(),
   cardsLeft: new Deck(),
+  cardsBottom: [] as number[],
   // Info
   player: {} as InternalPlayer,
   players: [] as PlayerInfo[],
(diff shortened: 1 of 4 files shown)


Manuel-777:
Related issue: #419

Thaoden:
At what point in the deck should they be re-added to the calculation though? when you cycled through the whole deck?
What happens when the deck is shuffled?

Manuel-777:
> At what point in the deck should they be re-added to the calculation though? when you cycled through the whole deck?
> What happens when the deck is shuffled?

All of those mentioned situations would work as you would expect, since we are not reacting to annotation events or things like that, we are reading trough the library game objects and finding the ones on the bottom, then removing those in the bottom from the calculation. When you shuffle the library for example, the game objects list from the library is updated and the cards on bottom get updated.

Thaoden:
Fair enough. Also, the case I made with the cycled deck is a non-case, as by then you have full knowledge of the next card...

PR #1037, opened 2020-04-27
Adjust optional property

Thaoden:
Just a miniscule fix adjusting the type according to [this line](https://github.com/Manuel-777/MTG-Arena-Tool/blob/548031d6b742444115a93f0a839809ad045cb8b2/src/window_background/greToClientInterpreter.ts#L657) - mark the property as maybe undefined.


Manuel-777:
Yeeeah.. thats the file we cannot touch.

Thaoden:
Ah... again :D

PR #1038, opened 2020-04-27
Fix negative wildcard input

### LINKED ISSUE #994: Explore tab wildcards filter allow negative numbers
https://discordapp.com/channels/463844727654187020/463844728287395851/700728730091388988

I don't think this is intended.

Thaoden:
This attempts to fix #994.

This introduces an optional property `onChange` that can be set on custom `Input` elements. If it is set, it will be executed before the original hook logic. The property is optional on the one hand to not break existing implementations, on the other hand because it seems not to be necessary in most cases.

With `onChange` in place, the wildcard input fields on the explore tab can validate their values and fall back to an empty value in case a non-valid value (read here: a negative value) has been entered.

### DIFF
--- a/src/window_main/components/misc/Input.tsx
+++ b/src/window_main/components/misc/Input.tsx
@@ -10,7 +10,7 @@ interface InputProps {
   title?: string;
   autocomplete?: string;
   callback?: (value: string) => void;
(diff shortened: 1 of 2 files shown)


Thaoden:
Naming is hard.
The way it is currently used, what the optional `onChange` really does is validating the input. So it should rather be named `validate`.

PR #1040, opened 2020-04-28
Cards seen by game

Manuel-777:
Displays cards seen separated by game number in match history view.

It is coded to work only after `v4.1.0`

image


riQQ:
Will there also be a combined view?

Manuel-777:
> Will there also be a combined view?

Obviously not yet!

Issue #1078, opened 2020-05-19
Upcoming changes with the new client version on 21.05.2020


riQQ:
> We're doing some behind-the-scenes work, which means there will be two changes to your Output Logs. First, it's super important to note that the file name is changing from output_log.txt to Player.log. You'll also have a convenient Player-prev.log as well, but it only saves one previous copy. The file itself will be saved in the same location (e.g. . . . \AppData\LocalLow\Wizards of The Coast\MTGA\Player.log).

See https://magic.wizards.com/en/articles/archive/magic-digital/mtg-arena-state-game-may-2020-05-19

PR #1079, opened 2020-05-19
New log path

Manuel-777:
In prep for new changes! #1078 

The popup should not be necessary if the old `output_log.txt` is deleted, but I can see that not happening, we should wait and see. It should be a smooth ride overall.


Thaoden:
What about using a fallback value, as not all users may update the tool and the game at the same time? Look for `Player.log`, if it isn't there, look for `output_log.txt`, if that isn't there, show an error.

Manuel-777:
As it is now, it will use default path to the new one (as it should be)

For players that have used the tool before the log path will be stored in the settings and will use that path, but if the new `Player.log` is also found it will ask them to change it.

As it is now, if you havent updated it will work with the path in settings (probably output_log)

PR #1080, opened 2020-05-20
Fix On the Play text: it was always Draw for g1

Porkepix:
This PR fix a probably wrong test for caption text in History for game 1: in every situation it displayed the "On the Draw" text.

I also would expect for `g2OnThePlay` and `g3OnThePlay` variables to not exist in case of BO1, but to be honest I am absolutely not knowledgeable of Node, JS and TypeScript, so there might be perfectly good reasons to declare and set values for those variables that I just don't understand.

### DIFF
--- a/src/window_main/components/misc/ResultDetails.tsx
+++ b/src/window_main/components/misc/ResultDetails.tsx
@@ -46,7 +46,7 @@ export default function ResultDetails(props: ResultDetailsProps): JSX.Element {
   let g1Title;
   if (match.gameStats[0]) {
     g1Title =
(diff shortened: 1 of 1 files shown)


Manuel-777:
Nice catch! The others are okay because we actually check for the games to exist before declaring the values, and let them lazy defined (`let g2Title;`) so we do not check against the title, but against the data itself when we need to access it.

Porkepix:
> Nice catch! The others are okay because we actually check for the games to exist before declaring the values, and let them lazy defined (`let g2Title;`) so we do not check against the title, but against the data itself when we need to access it.

Okay, from my understanding, g2 and g3 ones were always declared and set, but I don't master the language at all.

PR #1084, opened 2020-05-22
Fix log file selector default extension filter

Porkepix:
It was defaulting at looking for Text files with specific extensions,
therefore resulting for Player.log file being invisible to user.

---
Screenshot of the issue:
player_log_1

I was unsure of how to label this, I simply picked "Log", but it could also have been "Log File", "Log file", "Log Files" or "Log files".

Also, I'm unable to understand the logic of such a huge codebase, especially with languages I don't know and understand to answer the question myself, but couldn't those two part I modified be a single function called two times? It looks – from far – as duplicated code here, but again I have no knowledge nor understanding/skill here.

### DIFF
--- a/src/window_main/components/popups/OutputLogInput.tsx
+++ b/src/window_main/components/popups/OutputLogInput.tsx
@@ -38,7 +38,7 @@ export default function OutputLogInput(
         defaultPath: logUri,
         buttonLabel: "Select",
         filters: [
-          { name: "Text", extensions: ["txt", "text"] },
+          { name: "Log", extensions: ["log"] },
           { name: "All Files", extensions: ["*"] }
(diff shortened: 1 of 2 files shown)


Porkepix:
(Damn, just noticed that… I wrote a different Filter text in the two locations… ; tell me which one do you prefer so that I fix the commit, or you might want to fix the text yourselves)

Manuel-777:
You're right about the duplicate code there! Probably a leftover from older versions. Ill make sure to unify those later!

Porkepix:
@Manuel-777 Wouldn't have it been better to uniformize filter's text? (What I saw later that I mistakenly did set a different text).

PR #1089, opened 2020-05-23
Fix missing pip on first card

### LINKED ISSUE #1082: "New card" pip not shown on first card
The pips for opened or otherwise new cards currently don't show the first of a given card. For subsequent cards, this works fine.

Capture

Thaoden:
This attempts to fix #1082 .

It reuses the logic already present and extracts it into a helper function. A new card now shouldn't have a count of `NaN` anymore, but it should always get set to `1`.

It also introduces some new types for typing the reducers more explicitly.

`GreTypes` wasn't touched, these changes were introduced by running `npm run format`.


Thaoden:
Based on `InPlayerInventoryGetPlayerInventory`, `boosters` is of type `number[]`. `EconomyHeader.tsx` and `collectionStats.ts` had the type set as `boosters: {idx:number}[]`. The latest commit streamlined this.

However, atm I can't check if this still works as intended, as I don't have any boosters in my inventory.

PR #1090, opened 2020-05-23
Reducer types

Thaoden:
Just a bunch of types.

Next step is typing the possible `action`s for the reducers themselves.


src/shared-redux/slices/rendererSlice.ts
```diff
@@ -50,6 +54,11 @@ const initialRendererState = {
 };
 
 type RendererState = typeof initialRendererState;
+type Patreon = typeof initialRendererState.patreon;
+type Popup = typeof initialRendererState.popup;
```
Manuel-777, line 58: Uses of this can be replaced with `initialRendererState["popup"]` instead of declaring a new type (ditto below lines)

src/shared-redux/actions.ts
```diff
@@ -1,92 +1,167 @@
...
+  | "SET_DRAFT"
+  | "SET_MANY_DRAFT"
+  | "SET_SEASONAL"
+  | "SET_MANY_SEASONAL"
+  | "SET_DECK_CHANGE"
+  | "SET_MANY_DECK_CHANGES";
+
+const actions: Record<
+  Actions,
+  ActionCreatorWithPayload<any, string> //<P, T> where P is the type of the payload and T the type of the Action
```
Manuel-777, line 93: I'm not sure if its possible to make `reduxAction()`'s `arg` argumnet to inherit the types defined in the slices themselves? so when we write the actions these are smart enough to know if the action passed is correctly typed.

Thaoden:
The latest version (2f6c502) _does not transpile_!! It is more of a backup of the current state.

However, this version basically has all the types ready for use. Some types might have to be reworked to better align with the existing usages.

src/overlay/OverlayController.tsx
```diff
@@ -162,6 +166,10 @@ export default function OverlayController(): JSX.Element {
 
     reduxAction(
       store.dispatch,
+      {
+        type: "SET_SETTINGS",
+        arg: { overlays: newOverlays }
```
Manuel-777, line 171: Hm, so the arguments need to be defined twice now?

Thaoden: No, this is just a reminder for me what the original version looked like. Once I squish all the compile errors, I'll remove the third and fourth args so that the signature will look like
```TypeScript
reduxAction(
  store.dispatch, 
  {
    type:"SET_SETTINGS", 
    arg: {overlays: newOverlays}
  }, 
  IPC_ALL ^ IPC_OVERLAY 
);
```

Thaoden:
Made obsolete by porting this onto the another-webpack-try branch

PR #1102, opened 2020-05-30
Fix date type and format

dougmill:


src/window_background/loadPlayerConfig.ts
```diff
@@ -126,7 +126,9 @@ export async function loadPlayerConfig(): Promise<void> {
         savedData[id].playerDeckHash = playerDeck.getHash();
       }
 
-      savedData[id].date = new Date(savedData[id].date).toString();
+      savedData[id].date = savedData[id].date
+        ? new Date(savedData[id].date).toISOString()
+        : new Date(0).toISOString();
```
Thaoden, line 131: What's the resoning behind using `Date(0)` and not eg `Date.now`?

Thaoden: Also, it might be possible (didn't try) to rewrite this as
```TypeScript
savedData[id].date = new Date(savedData[id].date ?? 0).toIsoString();
```

dougmill: If a match somehow does not have a date (one of mine does), I don't think it should be included in anything that cares about any particular time span. `Date.now` would break that. Note that this is loading from the db file, so there is no reason to expect that the match is recent.

Thaoden: Fair enough

Thaoden:
Can the changes from `toString` to `toISOString` affect anything downstream?

dougmill:
> Can the changes from `toString` to `toISOString` affect anything downstream?
The initial value when the match or economy change is first saved has always been set with `toISOString`, so any code that might have to handle that has been handling it all along.

PR #1103, opened 2020-05-30
Don't ignore old matches

dougmill:


src/window_background/loadPlayerConfig.ts
```diff
@@ -105,30 +105,23 @@ export async function loadPlayerConfig(): Promise<void> {
...
   // Get Matches data
-  const newMatchesIndex: string[] = [];
-  Object.keys(savedData).forEach(k => {
-    if (k.indexOf("-" + arenaId) !== -1) {
-      newMatchesIndex.push(k);
+  const newMatchesIndex: string[] = Object.keys(savedData).filter(
+    id =>
+      savedData[id] &&
+      savedData[id].gameStats?.length > 0 &&
+      savedData[id].gameStats[0] !== undefined
```
Thaoden, line 112:
```suggestion
id => savedData[id]?.gameStats?[0] !== undefined
```

src/window_background/loadPlayerConfig.ts
```diff
@@ -106,7 +106,7 @@ export async function loadPlayerConfig(): Promise<void> {
 
   // Get Matches data
   const newMatchesIndex: string[] = Object.keys(savedData).filter(
-id => savedData[id]?.gameStats?[0] !== undefined
+    id => savedData[id]?.gameStats?.[0] !== undefined
```
Thaoden, line 109: Not sure about the dot in `gameStats?.[0]`, you don't index after dotting, normally.

dougmill: It's the syntax for a null-aware indexing operator. Without the dot, it's parsed as a ternary and gets an error because the corresponding colon isn't there.

Thaoden: TIL. thanks!

PR #1105, opened 2020-05-31
Another webpack try

Manuel-777:


Thaoden:
The `.vscode` folder contains local workspace settings - these shouldn't be checked in to source control IMHO.

PR #1106, opened 2020-06-01
Re-push match data that was lost on the server

dougmill:


src/window_background/httpApi.ts
```diff
@@ -265,18 +281,26 @@ function handleSync(syncIds: SyncIds): void {
...
+}
+
+function handleRePushLostMatchData() {
+  const toPush: SyncIds = {
+    courses: [],
+    matches: Object.keys(globalStore.matches).filter(id => {
+      const match = globalStore.matches[id];
+      return (
+        !match.lastPushedDate &&
+        "2019-01-28T00:00:00.000Z" < match.date &&
```
Thaoden, line 294: What do the "magic dates" represent?
They're redundant though, if the first is true, the second has to be true.

src/window_background/httpApi.ts
```diff
@@ -500,11 +526,20 @@ function handleSetDataResponse(
 ): void {
   const mongoDbDuplicateKeyErrorCode = 11000;
   finishSync();
-  if (parsedResult && parsedResult.error === mongoDbDuplicateKeyErrorCode) {
-    return; // idempotent success case, just return
-  } else if (error) {
+  // duplicate key is idempotent success case, don't count it as error
+  if (
+    !(parsedResult && parsedResult.error === mongoDbDuplicateKeyErrorCode) &&
```
Thaoden, line 531:
```suggestion
parsedResult?.error !== mongoDbDuplicateKeyErrorCode &&
```

src/window_background/httpApi.ts
dougmill: This is for recovering data lost in a specific incident, not a general recovery algorithm, and that's approximately the date range that data was lost for.

Thaoden: Ah nevermind the redundancy, the variable position changed.
As a nit: I'd change the operator instead of the variable positioning for the second date check. Also, if possible, give names to magic values.
Not code related, but... Recovering data from a year ago?

dougmill: I want it for the new shuffler data presentation page I'm making. That time span happens to include the entire period from when shuffler data collection started to when the bug was fixed, so without it I can't show the dramatic change that happened with the WAR release.

Thaoden: Something along the lines of `const shufflerDataCollectionStart: string = "2019-01-28T00:00:00.000Z";
const shufflerBugFixDate: string = "2019-05-01T00:00:00.000Z";` ?

dougmill: >As a nit: I'd change the operator instead of the variable positioning for the second date check.

That would be much less noticeable to me.

Thaoden: Depends on what you expect and what you're used to, I guess ;)

Thaoden: I'd define the constants outside of the filter to avoid redefining them.

PR #1112, opened 2020-06-04
Tests for bounds

Manuel-777:


src/main/getNewBounds.ts
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable @typescript-eslint/camelcase */
+import electron from "electron";
+
+export default function getNewBounds(): electron.Rectangle {
+  const newBounds = { x: 0, y: 0, width: 0, height: 0 };
+  electron.screen.getAllDisplays().forEach((display) => {
+    newBounds.x = Math.min(newBounds.x, display.bounds.x);
+    newBounds.y = Math.min(newBounds.y, display.bounds.y);
+  });
```
AnnanFay, line 9: I'd write this as something like:

```javascript
  let displays = electron.screen.getAllDisplays();
  newBounds.x = Math.min(...displays.map(display => display.bounds.x));
  newBounds.y = Math.min(...displays.map(display => display.bounds.y));
```

Not sure which is more performant, but this makes it clear we want the minimum of all the x values of all the displays. And the same for y values. Nothing wrong with the current code though, just found it interesting.

PR #1113, opened 2020-06-08
Human drafts

Manuel-777:
Added human drafts, but then something happened.

It turns out this was a nice excuse for me to revamp out internal drafts structure, using arrays and a proper compact interface that does everything we need. Unfortunately the interface is too much trouble to keep alongside our old datasets for compatibility, so I rather ignore the old data entirely and use the new one.

I might include a converter but it might be too much trouble? Not sure.


src/background/arena-log-decoder/__tests__/arena-log-decoder-spec.js
```diff
@@ -17,7 +17,7 @@ describe("arena-log-decoder", () => {
       const decoder = ArenaLogDecoder();
       const logEntries = [];
       decoder.append(text, logEntry => logEntries.push(logEntry));
-      expect(logEntries.length).toEqual(19);
+      expect(logEntries.length).toEqual(37);
```
Thaoden, line 20: Is it possible to get the `37` value directly from `text` somehow?

src/background/onLabel/InDraftDraftStatus.ts
```diff
@@ -14,13 +17,12 @@ export default function InDraftDraftStatus(entry: Entry): void {
   if (!json) return;
 
   startDraft();
+  const pack = json.PackNumber;
+  const pick = json.PickNumber;
+  const currentPack = (json.DraftPack || []).slice(0).map((n) => parseInt(n));
```
Thaoden, line 22: Would it be possible to set `[]` as default value for `json.DraftPack`? Meaning it cannot be `undefined` anymore but is at least `[]` and this check becomes obsolete.

src/background/onLabel/InEventJoin.ts
```diff
@@ -3,9 +3,24 @@ import { InternalDeck } from "../../types/Deck";
...
 
 interface EntryJson {
-  CourseDeck: InternalDeck;
+  Id: string;
+  InternalEventName: string;
+  PlayerId: string | null;
+  ModuleInstanceData: ModuleInstanceData;
+  CurrentEventState: string;
+  CurrentModule: string;
+  CardPool: null | [];
```
Thaoden, line 21: What's the reasoning behind this being `null`?

src/background/onLabel/InEventJoin.ts
```diff
@@ -3,9 +3,24 @@ import { InternalDeck } from "../../types/Deck";
...
+  resetCurrentDraft,
+} from "../../shared/store/currentDraftStore";
+import { ModuleInstanceData } from "../../types/event";
+import getSetCodeInEventId from "../../shared/utils/getSetInEventId";
 
 interface EntryJson {
-  CourseDeck: InternalDeck;
+  Id: string;
+  InternalEventName: string;
+  PlayerId: string | null;
```
Thaoden, line 17: Is it possible to set a default value instead of this being `null`?

src/overlay/DraftElements.tsx
```diff
@@ -28,7 +28,7 @@ export interface DraftElementsProps {
  */
 export default function DraftElements(props: DraftElementsProps): JSX.Element {
   const { draft, draftState, index, setDraftStateCallback, settings } = props;
-  const packSize = packSizeMap[draft.set] || 14;
+  const packSize = packSizeMap[draft.draftSet] || 14;
```
Thaoden, line 31: What does the magical number `14` represent?

src/overlay/DraftElements.tsx
```diff
@@ -51,40 +51,35 @@ export default function DraftElements(props: DraftElementsProps): JSX.Element {
...
 
   const { packN, pickN } = draftState;
-  const isCurrent = packN === draft.packNumber && pickN === draft.pickNumber;
+  const isCurrent = packN === draft.currentPack && pickN === draft.currentPick;
   let visibleDeck = null;
   let cardsCount = 0;
   let mainTitle = "Overlay " + (index + 1);
   let subTitle = "";
-  let pack = [];
+  let pack = [] as number[];
```
Thaoden, line 74:
```suggestion
    let pack: number[] = [];
```

src/renderer/components/list-item/ListItemEvent.tsx
```diff
@@ -26,19 +26,20 @@ import RoundCard from "../misc/RoundCard";
...
+import { useSpring, animated } from "react-spring";
 
 function DraftRares({ event }: { event: EventTableData }): JSX.Element {
   const draftId = event.id + "-draft";
   let draftRares: JSX.Element[] = [];
   if (draftExists(draftId)) {
     const draft = getDraft(draftId);
-    if (draft && draft.CardPool) {
-      const pool = [...draft.CardPool];
+    if (draft && draft.pickedCards) {
```
Thaoden, line 41:
```suggestion
        if (draft?.pickedCards) {
```

src/shared/store/currentDraftStore.ts
```diff
@@ -0,0 +1,82 @@
...
+  ];
+}
+
+export function setDraftPack(
+  cards: number[],
+  argPack: number | undefined,
+  argPick: number | undefined
+): void {
+  const pack =
+    argPack == undefined ? globalStore.currentDraft.currentPack : argPack;
```
Thaoden, line 61:
```suggestion
const pack = argPack ?? globalStore.currentDraft.currentPack;
```

src/shared/store/currentDraftStore.ts
```diff
@@ -0,0 +1,82 @@
...
+  const pick =
+    argPick == undefined ? globalStore.currentDraft.currentPick : argPick;
```
Thaoden, line 63:
```suggestion
const pick = argPick ?? globalStore.currentDraft.currentPick;
```

src/shared/store/currentDraftStore.ts
```diff
@@ -0,0 +1,82 @@
...
+  globalStore.currentDraft.currentPack = pack;
+  globalStore.currentDraft.currentPick = pick;
+  globalStore.currentDraft.packs[pack][pick] = cards;
+}
+
+export function addDraftPick(
```
Thaoden, line 69: This is almost completely duplicated in `setDraftPack` above. Maybe refactor out the common functionality.

src/shared/store/index.ts
```diff
@@ -183,18 +185,18 @@ export function transactionsList(): InternalEconomyTransaction[] {
...
-  if (!id || !globalStore.drafts[id]) return undefined;
-  return globalStore.drafts[id];
+export function getDraft(id: string): InternalDraftv2 | undefined {
+  if (!id || !globalStore.draftsv2[id]) return undefined;
+  return globalStore.draftsv2[id];
 }
 
 export function draftExists(id: string): boolean {
-  return globalStore.drafts[id] ? true : false;
+  return globalStore.draftsv2[id] ? true : false;
```
Thaoden, line 194:
```suggestion
return globalStore.draftsv2[id] !== undefined;
```
Not sure if it can be null, too, though.

src/shared/store/index.ts
```diff
@@ -183,18 +185,18 @@ export function transactionsList(): InternalEconomyTransaction[] {
 //
 // Draft utility functions
 //
-export function getDraft(id: string): InternalDraft | undefined {
-  if (!id || !globalStore.drafts[id]) return undefined;
-  return globalStore.drafts[id];
+export function getDraft(id: string): InternalDraftv2 | undefined {
+  if (!id || !globalStore.draftsv2[id]) return undefined;
+  return globalStore.draftsv2[id];
```
Thaoden, line 190:
```suggestion
return id ? globalStore.draftsv2[id] : undefined;
```

src/types/draft.ts
```diff
@@ -30,6 +31,22 @@ export interface PickPack {
   pack: string;
 }
 
+export interface InternalDraftv2 {
+  archived: boolean;
+  owner: string;
+  arenaId: string;
+  date: string;
+  eventId: string;
+  id: string | undefined;
```
Thaoden, line 40: Is it possible to use a default value instead of setting it `undefined`?

src/background/arena-log-decoder/__tests__/arena-log-decoder-spec.js
Manuel-777: The idea is that when you fix/change the log decoder these values match (it was 33 initially when Gnombre coded it but I messed up a little bit and now its ok again because I added more labels)

src/background/onLabel/InDraftDraftStatus.ts
Manuel-777: Not sure, in the logs the type of it can be either `null` or `[]` so this check is needed

src/background/arena-log-decoder/__tests__/arena-log-decoder-spec.js
Thaoden: yeah, the intent is clear ;) but I think it a bit cumbersome to always have to adjust the "expected" value by hand when it should only depend on the input and so could be derived directly from it.

src/background/onLabel/InEventJoin.ts
Manuel-777: It would be an event joined but without any cards yet, like when you first join a sealed or draft event

Thaoden: So that's the same as `[]` then - no cards?

src/background/onLabel/InEventJoin.ts
Manuel-777: its the types I found in the logs, so no..

Thaoden: Fair enough

src/overlay/DraftElements.tsx
Manuel-777: `DEFAULT_PACK_SIZE`

src/background/onLabel/InEventJoin.ts
Manuel-777: yep!

src/types/draft.ts
Manuel-777: I dont want a default value to slip trough and get into the db, id go with the `?`

src/shared/store/index.ts
Manuel-777: not sure, but it doesnt hurt to leave it ?

Thaoden: No it doesn't. It's just that `return condition ? true : false;` always look funny to me, when you could just `return condition;` ;)

src/background/arena-log-decoder/__tests__/arena-log-decoder-spec.js
Manuel-777: Actually this value *should not* be changed or change unless we touch the log decoder in a way that possibly breaks it, like it happened in this case.

Manuel-777:
Added old drafts converter, this seems g2g for me! @riQQ @Thaoden

src/shared/utils/convertDraftToV2.ts
```diff
@@ -0,0 +1,57 @@
...
+      getSetCodeInEventId(original.InternalEventName) ||
+      "",
+    currentPack: original.packNumber,
+    currentPick: original.pickNumber,
+    // removed number from the original CardPool to make this work
+    // Basically we cant map a string[] | number[] array and get it
+    // typed as a number[] automatically based on the map fn
+    pickedCards: original.CardPool?.map((n: string | number) =>
+      parseInt(n + "")
+    ) as number[],
```
Thaoden, line 52: This might break if `CardPool` is `undefined` but the typing says it's a `number[]`, maybe add a check for `undefined`?

src/background/loadPlayerConfig.ts
```diff
@@ -203,6 +204,20 @@ export async function loadPlayerConfig(): Promise<void> {
     );
   }
 
+  // Get old drafts data and convert
+  if (savedData.draft_index) {
+    const draftsList: InternalDraftv2[] = savedData.draft_index
+      .filter((id: string) => savedData[id])
+      .map((id: string) => convertDraftToV2(savedData[id] as InternalDraft));
```
Thaoden, line 211: Why the cast to `InternalDraft` when the fn returns a `InternalDraftv2`?

Issue #1117, opened 2020-06-14
Failure to load local user DB due to bad draft format


tkuneck:
Version: 5.0.5

I had data in my user.db from a draft that did not have the InternalEventName field specified. It instead was only present in the associated type:event record in the DB. When loading, I debugged to where the new draft logic is attempting to upgrade from InternalDraft -> InternalDraftv2. There is a call via convertDraftToV2 sending the InternalDraft.InternalEventName to the new getSetCodeInEventId function, which throws on receiving undefined and trying to call .indexOf on the undefined object.

This blocked loading the tool, which just sat spinning due to the backend throwing in getSetCodeInEventId.

I've fixed this locally, though I don't think this is the optimal solution. It handles the offending data in my user.db:
https://github.com/tkuneck/MTG-Arena-Tool/commit/7456326f93fc466675445a95efab285b980eb811

I've built this and have it running locally, everything seems fine.

I'm attaching the rows of my user.db that triggered the failure.
[problemdata.json.gz](https://github.com/Manuel-777/MTG-Arena-Tool/files/4775774/problemdata.json.gz)

tkuneck:
Fixed by commit: https://github.com/Manuel-777/MTG-Arena-Tool/commit/acbecebe7ae37b4fc7b08801c633c8c23cd52fda

PR #1118, opened 2020-06-14
Fix unknown internal event name

tkuneck:
A fix for: https://github.com/Manuel-777/MTG-Arena-Tool/issues/1117

Add some fall-back logic to identify draft details when upgrading from InternalDraft to InternalDraftV2, and add some resiliency to undefined input to the getSetCodeInEventId function.

### DIFF
--- a/src/shared/utils/getSetInEventId.ts
+++ b/src/shared/utils/getSetInEventId.ts
@@ -1,17 +1,17 @@
 import database from "../database";
 
 export default function getSetCodeInEventId(
-  eventId: string
+  eventId: string | undefined
 ): string | undefined {
-  if (eventId) {
-    const setCodes = Object.keys(database.sets)
-      .filter(
-        (setName) => eventId.indexOf(database.sets[setName].arenacode) !== -1
-      )
-      .map((setName) => database.sets[setName].arenacode);
-
-    return setCodes[0] || undefined;
+  if (!eventId) {
+    return undefined;
   }
 
-  return undefined;
+  const setCodes = Object.keys(database.sets)
+    .filter(
+      (setName) => eventId.indexOf(database.sets[setName].arenacode) !== -1
+    )
+    .map((setName) => database.sets[setName].arenacode);
+
+  return setCodes[0] || undefined;
 }
--- a/src/shared/utils/getSetInEventId.ts
+++ b/src/shared/utils/getSetInEventId.ts
@@ -4,13 +4,13 @@ export default function getSetCodeInEventId(
   eventId: string
 ): string | undefined {
   if (eventId) {
-      const setCodes = Object.keys(database.sets)
-        .filter(
-          (setName) => eventId.indexOf(database.sets[setName].arenacode) !== -1
-        )
-        .map((setName) => database.sets[setName].arenacode);
+    const setCodes = Object.keys(database.sets)
+      .filter(
+        (setName) => eventId.indexOf(database.sets[setName].arenacode) !== -1
+      )
+      .map((setName) => database.sets[setName].arenacode);
 
-      return setCodes[0] || undefined;
+    return setCodes[0] || undefined;
   }
 
   return undefined;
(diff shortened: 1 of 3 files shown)


src/shared/utils/getSetInEventId.ts
```diff
@@ -3,11 +3,15 @@ import database from "../database";
 export default function getSetCodeInEventId(
   eventId: string
 ): string | undefined {
-  const setCodes = Object.keys(database.sets)
-    .filter(
-      (setName) => eventId.indexOf(database.sets[setName].arenacode) !== -1
-    )
-    .map((setName) => database.sets[setName].arenacode);
+  if (eventId) {
```
Thaoden, line 6: If `eventId` can be `undefined` here, it should be reflected in the function signature.

Additionally, I'd revert the logic, `if (!eventId) return undefined;` but that is a matter of taste, I guess.

Thaoden:
If I got your diagnosis right, the problem lies in `src/background/loadPlayerConfig.ts` line 213 probably passing in `undefined`? I would then rather check on the calling side if the argument I'm going to pass in is defined and, if not, pass in the alternative argument, instead of passing in two arguments and let the callee decide which one to use.

src/shared/utils/getSetInEventId.ts
tkuneck: ACK on updating the function signature, will change.

I originally checked for undefined and returned null, thought this might be cleaner. I don't normally write TS code, I'm less clear on the common patterns for handling things like undefined checks. My google searching led me to this conditional execution only if eventId is defined due - seemed safer.

tkuneck:
Yes, 213 is passing undefined. Checking on the calling side seemed much more painful to implement since I found the required data in another DB entry. Allowing the downstream logic to pick up and use the alternative data as appropriate seemed better in this case as there were multiple places where missing data could be filled in (InternalEventName and CardPool). The problem is that loadPlayerConfig is passing a complex object (InternalDraft) and data from another complex object (also InternalDraft) has 2+ datapoints to merge into what is passed

If this were handled in loadPlayerConfig, where would you suggest the data from the two InternalDraft objects gets merged? A new merging function perhaps?

Manuel-777:
Great job on this!

I think it might be easier to check on ´L213´ directly then pass the checked or fixed data onto ´convertDraftToV2()´, something like
´´´
.map(id: string) => {
  const metadata = savedData[id.replace(/-draft$/g, "")];
  if (!savedData[id].InternalEventName && metadata?.InternalEventName) {
    savedData[id].InternalEventName = metadata.InternalEventName;
  }
  convertDraftToV2(savedData[id] as InternalDraft);
}):
´´´

This way you dont have to add an aditional parameter to ´convertDraftToV2()´ and check it all the way in. Altrough I agree adding a check for ´undefined´ would be nice.

src/shared/utils/getSetInEventId.ts
tkuneck: Changed signature, and changed to if (!eventId) return undefined;

Thaoden:
Looks good to me!

tkuneck:
There are two options:
1. apply the fix globally as part of the call to fixBadPlayerData (implemented at commit db1c656eaf1c5ceb5f6ee6ae8b53318d8e88d0e4)
2. apply the fix locally to the conversion to v2 in the draft_index processing (implemented at commit f9e29950cbcffb8c7289d9b77a819dc952891bbc)

I kinda like option 2 now that it's implemented, but I'm happy to revert and go back to option 1 if the global implementation is preferred.

tkuneck:
So it's easier to observe, diff for option 1:
https://github.com/Manuel-777/MTG-Arena-Tool/compare/ac0e99bd..db1c656eaf1c5ceb5f6ee6ae8b53318d8e88d0e4

diff for option 2:
https://github.com/Manuel-777/MTG-Arena-Tool/compare/ac0e99bd..f9e29950cbcffb8c7289d9b77a819dc952891bbc

PR #1125, opened 2020-06-25
Update proto for patch version 1.09.00.00

riQQ:


Manuel-777:
Ran the GreTypes update, and tests look good!

PR #1143, opened 2020-07-12
Sidebar Matchup Matrix using local aggregation data system

AnnanFay:
The goal is to use local data to provide a useful matchup table with minimal clutter. Once a person is familiar with the table construction it should easily provide useful data by glancing at it / scanning.

It's integrated into the aggregation system and placed in the sidebar so that any of the current filtering methods will work. The below is filtered to ranked constructed.

image

Floating over cells shows more information and highlights the row and column:
image

The data is sorted in the same method as the other sidebar charts - by frequency. Shape and size of tiles gives information of sample size for that matchup with roughly exponential scaling.

The colour scheme is viridis from matplotlib, details [here](https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html). As future work there should be a setting to switch between different colour palettes. Colour blind simulation of green-red palette versus viridis:
image

There are issues with the archetype system which are beyond the scope of this pull request and should be fixed separately. The biggest problem is that tags are asymmetric. Opponents decks are tagged with archetype but your decks are tagged with format. Thus there is no easy way of doing a proper archetype vs archetype matchup.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4978,6 +4978,14 @@
       "resolved": "https://registry.npmjs.org/color-name/-/color-name-1.1.3.tgz",
       "integrity": "sha1-p9BVi9icQveV3UIyj3QIMcpTvCU="
     },
+    "colormap": {
+      "version": "2.3.1",
+      "resolved": "https://registry.npmjs.org/colormap/-/colormap-2.3.1.tgz",
+      "integrity": "sha512-TEzNlo/qYp6pBoR2SK9JiV+DG1cmUcVO/+DEJqVPSHIKNlWh5L5L4FYog7b/h0bAnhKhpOAvx/c1dFp2QE9sFw==",
+      "requires": {
+        "lerp": "^1.0.3"
+      }
+    },
     "colour": {
       "version": "0.7.1",
       "resolved": "https://registry.npmjs.org/colour/-/colour-0.7.1.tgz",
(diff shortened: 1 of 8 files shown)


Manuel-777:
Archetypes/tags have always been unreliable, and there is also the issue you mention with player tags being different than the ones used for the opponent's (they have different uses), shouldn't we go with the most straightforward comparisons between the two? Either Decks (by id/hash) vs Colors or Colors vs Colors chart?

Looking at my data now is like this, which is pretty much of no use to me, because some opponent tags are entirely wrong and because I use my tags to separate by format;
image

AnnanFay:
> shouldn't we go with the most straightforward comparisons between the two? Either Decks (by id/hash) vs Colors or Colors vs Colors chart?

You don't think it's good enough to just use the toggle at the top of the screen? That already gives a Colors vs Colors chart and is the main way I've been using this.

One thing I am going to do is change how the tag matrix works so it does work for people who manually set deck tags. We can loop over all tags assigned to a deck so that manually set tags appear.

The tag issue is worse than I thought after some digging. Issues:

1. Matches store the deck data from when the match happened, so you cannot retrospectively tag a match.
2. Deck tags are completely ignored when storing a match.
3. Opponent deck tags are stored in the match with the actual deck `.tags` being [null].

To test the issue I manually tagged a brawl deck as Yorvo, Lord of Garenbrig. I played 1 game in Brawl. The match data has:
```javascript
match.playerDeck.colors = [5]
match.playerDeck.tags = ["Brawl"]
match.playerDeck.format = ""

match.oppDeck.colors = [1,2,4,5]
match.oppDeck.tags = [null]
match.oppDeck.format = ""

```

PR #1145, opened 2020-07-15
[feature] Jumpstart

Manuel-777:
Preliminary PR to support Jumpstart boosters theme detection.

- [x] Detect themes based on deck list.
- [x] Test decks theme detection.
- [x] Add themes as tags on match end.
- [x] Add additional object key in match data to help with aggregation of themes.


Manuel-777:
used [mtg.gamepedia.com/Jumpstart](https://mtg.gamepedia.com/Jumpstart) to make the cards/theme map, Im not sure if some or any of those cards are in the list of modified cards for MTG Arena though!

PR #1153, opened 2020-07-23
Fix uploading of Bo3 matches

dougmill:


src/types/match.ts
```diff
@@ -94,7 +94,7 @@ export interface InternalMatch {
   oppDeck: InternalDeck;
   tags: any;
   date: string;
-  lastPushedDate: string;
+  lastPushedDate?: string; // not set until match is successfully pushed
```
dougmill, line 97: This was optional when originally added, and I have no idea why that was changed.

src/background/saveMatch.ts
```diff
@@ -89,8 +89,6 @@ function generateInternalMatch(
     toolVersion: globals.toolVersion,
     toolRunFromSource: !electron.remote.app.isPackaged,
     arenaId: globals.store.getState().playerdata.playerName,
-    // TS complains about undef but lastPushedDate should not be declared here
-    lastPushedDate: new Date().toISOString(),
```
dougmill, line 93: The issue commented about was caused by removing the `?` in the field definition. Setting it to the current date here negated the point of the field entirely, which is why I had to add `lastPushedByVersion` for an alternative.

PR #1191, opened 2020-11-02
Feature cards tooltip

eyeq:
Added a card image tooltip to Cards daily header on the Economy tab.
It uses the react-tooltip library.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13533,6 +13533,15 @@
         }
       }
     },
+    "react-tooltip": {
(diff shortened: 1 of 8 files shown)


Manuel-777:
I think there needs to bne a way tosee those cards, but the tooltip is just too hard to see, maybe a popup would be better? something we can hover on each card individually that is fixed on the screen.

eyeq:
When using the react-tooltip library, it looks like this.

PR #1192, opened 2020-11-10
Search tags the user has tagged when saveMatch

eyeq:
Added the feature to search archetypes (tags) from the client database.
This is a feature I personally want, so we may need to switch on/off in the setting option.


Manuel-777:
It looks good, why you think it should be turned off? Was it too slow to process?

eyeq:
I needed this feature, but I thought some people might not need it.

PR #1196, opened 2020-11-12
Added format-tag to DecksArtView

eyeq:


Manuel-777 (changes requested):
Not merging this one, it has a an issue when scrolling (see image below), but most importantly it just does not look good. I played before putting this information on the Black/Transparent box, but its too tight with data already it felt non important.
If you can find a way to make it look clean, not obstructing the art, etc, we can give it another try!

image

eyeq:
Oops, it's fine in my environment, but it seems that my debugging wasn't enough.
I will fix it.

キャプチャ

PR #1202, opened 2020-11-16
Update search tags the user has tagged

eyeq:
I don't know how much subdivision is best, 
but at least Sealed, Draft, and Brawl have stricter deck building restrictions than other formats, so I've found that archetypes in other formats aren't useful.


eyeq:
thanks

